PR #203, opened 2019-05-16
Fixed casing in link back to readme.md

dvdstelt:
Small thing that annoyed me and returned 404 pages :-)


LadyNaggaga:
Thanks for catching that 👍  because this hasn't happened in the past. I am glad you caught thanks
README

dvdstelt:
I'm trying to load the solution with Visual Studio **2019** to verify why the build fails, but it won't even load! VS2019 completely hangs on loading MLS.Agent.Tests.

jonsequitur:
@dvdstelt Thanks. The failed due to a flaky test.

Now as to why VS2019 hangs for you, that's a different question. You might try running `build.cmd` first.

PR #205, opened 2019-05-16
prevent samples from ingesting the arcade sdk

brettfo:
Running `dotnet run -- ..\Samples` from `MLS.Agent` was crawling up the directory tree and eventually was pulling in the dotnet internal build infrastructure, which was messing everything up.  This simply adds files that prevents the directory crawling and removes `Snippets.csproj` from the solution, since we don't want it build with the dotnet team infrastructure.


jonsequitur:
We've also been discussing moving sample and demo content to a different repo, which is a logical next step here.

@LadyNaggaga

PR #234, opened 2019-05-22
Build client and js api

colombod:

PR #237, opened 2019-05-25
initial support for F#

brettfo:
**tl;dr** - Editing/running F# snippets in Try .NET and displaying diagnostics.

Oh boy, is this a giant pile of hacks.

I'm creating this PR so early in the process to get feedback on how to better architect these changes, because right now I'm rather ashamed of what I've done.

List of sins and why I committed them (pun intended):

1. ~~Workspace servers: Rather than try to factor out all the calls to `RoslynWorkspaceServer` and properly abstract them, I created `WorkspaceServerMultiplexer` which simply looks for the suffix `.fsproj` on the workspace type to redirect to `FSharpWorkspaceServer`; otherwise it defaults to `RoslynWorkspaceServer`.  As I understand it, this will only work for on-disk workspaces which just happen to set their type to the project file path (see item 1 below).~~ **I've mostly plumbed through a `Language` property, and otherwise auto-detected it from the `.fsproj` suffix for the workspace servers.**
2. ~~In an attempt to reuse as much as possible (meaning less plumbing), I hijacked various functions to explicitly look for F#-specific file extensions and then bail in a different direction.  See `BufferInliningTransformer.cs`, `FileExtensions.cs`, and `SourceTextExtensions.cs`.  A method in `BufferInliningTransformer.cs` has been made `virtual` to allow for F#-specific overriding.~~
3. ~~Markdig doesn't understand code fences of the type `fsharp`, so to get it to appropriately classify F# blocks as `AnnotatedCodeBlock`s, I had to lie and say it was C#.  The proper fix for this should be adding F# support to [Markdig](https://github.com/lunet-io/markdig).~~  **Turns out it's not Markdig, but this repo; I found and updated the language parsing.**
4. F# can't perform compilations and analysis on purely in-memory data structures like Roslyn can, so I introduced `RedirectedPackage` which creates a directory under `%TEMP%`, copies everything from the parent `Package`, and does the text substitution and all work there.  (This also explains why I do source path mapping for F# diagnostics in `Library.fs`.)  After compilation the `bin/$(Configuration)` directory is then copied back to the 'real' location to allow execution.  See item 6.A below.

Things that still need to be done:

1. ~~Allow a workspace to properly state it's preferred server: `RoslynWorkspaceServer` or `FSharpWorkspaceServer`.~~
2. ~~Parameterize the hijacked methods and call sites to make the separation cleaner.  See item 2 above.~~ Cleaned up as much as possible for this PR.  Will be cleaned up even more when #254 is integrated.
3. Implement completion in `FSharpWorkspaceServer.cs` (future PR).
4. Implement signature help in `FSharpWorkspaceServer.cs` (future PR).
5. Add F# tests ~~and fix whatever other tests I most certainly broke.~~
6. (Long lead) update the F# libraries to allow in-memory compilation and analysis.
  A. In the short term we could likely hijack the call to `dotnet build` and pass an additional property, `/p:OutputPath=path\to\original\project\bin`.

Uninteresting note: The C# runner uses `#region`/`#endregion` tags to designate which parts of the file to display in the editor.  F# doesn't have the concept of regions but other tools cheat and use the specially-formatted comments, `//#region` and `//#endregion`, so I also opted for that approach.


Microsoft.DotNet.Try.Project/FSharpMethods.cs
```diff
@@ -0,0 +1,57 @@
...
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis.Text;
+using Microsoft.DotNet.Try.Protocol;
+using Buffer = Microsoft.DotNet.Try.Protocol.Buffer;
+
+namespace Microsoft.DotNet.Try.Project
+{
+    internal class FSharpMethods
+    {
+        private const string FSharpRegionStart = "//#region";
```
jonsequitur: >"//#region" [](start = 49, length = 11)

Is this ideal? Any suggestions for making it more natural in F#?

WorkspaceServer/Packaging/PackageInitializer.cs
```diff
@@ -58,6 +61,17 @@ public class PackageInitializer : IPackageInitializer
                 await afterCreate(directory, budget);
             }
         }
+
+        private static string GetLanguageFromProjectName(string projectName)
```
jonsequitur: >string [](start = 23, length = 6)

A more constrained type would be a good idea.

Microsoft.DotNet.Try.Project/FSharpMethods.cs
brettfo: Not really; F# doesn't have the equivalent of `#region` or any concept of a code block.  The closest thing we have is [computation expressions](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/computation-expressions), but that would require the writer of the documentation/sample to define a new builder for each span of code.  E.g., the backing code *could* look like this:

``` F#
type DemoRegionBuilder() =
    member __.Zero() = ()

// ... elsewhere
let demo_region_1 = new DemoRegionBuilder()
let demo_region_2 = new DemoRegionBuilder()
// .. repeat for every equivalent of `#region demo_region_X` that would be found in C#

[<EntryPoint>]
let main(args: string[]) =
    let _ =
        demo_region_1 {
            printfn "hello, world" // this line appears in the web page
        }
    0
```

With the markdown looking like this:

``` markdown
` ` ` fsharp --project ./MySampleProject.fsproj --source-file ./MySampleFile.fs --region demo_region_1
printfn "hello, world" // this line appears in the web page
` ` `
```

Adding @KevinRansom to see if he has any additional insight, but I'm still leaning towards specially-formatted comments, because that's essentially what C#/VB `#region`s are anyways.

WorkspaceServer/Packaging/PackageInitializer.cs
brettfo: If I'm understanding this correctly, once #234 is in we can modify the `WorkspaceRequest` object to contain a language name which would then allow the removal of the `WorkspaceServerMultiplexer` and by extension this method, because we'd then have access to the language at the `Workspace` level and it could then be plumbed through where appropriate.  E.g., if `WorkspaceRequest.Workspace.Language` was a valid value, then [this line](https://github.com/dotnet/try/blob/9e4d03378204b040dddfa0939b5a56fb7523ca49/MLS.Agent/Controllers/LanguageServicesController.cs#L130) would be able to appropriately select `RoslynWorkspaceServer` or `FSharpWorkspaceServer`.

FSharpWorkspaceShim/FSharpWorkspaceShim.fsproj
```diff
@@ -0,0 +1,23 @@
...
+    <PublishWindowsPdb>false</PublishWindowsPdb>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="Library.fs" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Update="FSharp.Core" Version="4.6.2" />
+    <PackageReference Include="FSharp.Compiler.Service" Version="28.0.0" />
```
cartermp: Note that version 29 is now out and does contain some breaking changes; might be good to just use that one here at the outset.

FSharpWorkspaceShim/Library.fs
```diff
@@ -0,0 +1,130 @@
...
+                    1
+                | c ->
+                    // handle remaining chars: control - replace with space, others - keep unchanged
+                    let c = if Char.IsControl c then ' ' else c
+                    buf.Append c |> ignore
+                    1
+            i <- i + delta
+        buf.ToString()
+
+    let private newlineifyErrorString (message:string) = message.Replace(newlineProxy, Environment.NewLine)
```
jonsequitur: "newlineify" 😆

MLS.Agent/Markdown/LocalCodeFenceAnnotationsParser.cs
```diff
@@ -113,7 +113,7 @@ private static void AddSourceFileOption(Command csharp)
                 var projectFiles = directoryAccessor.GetAllFilesRecursively()
                                                     .Where(file =>
                                                     {
-                                                        return directoryAccessor.GetFullyQualifiedPath(file.Directory).FullName == rootDirectory.FullName && file.Extension == ".csproj";
+                                                        return directoryAccessor.GetFullyQualifiedPath(file.Directory).FullName == rootDirectory.FullName && (file.Extension == ".csproj" || file.Extension == ".fsproj");
```
jonsequitur: There's a potential inconsistency here in that I can do this and it will be valid:

````markdown
```c# --project ./something.fsproj
```
````

Comment (brettfo) [reply MLS.Agent/Markdown/LocalCodeFenceAnnotationsParser.cs]: My plan (hope?) was to merge as-is to `feature/fsharp` (where this PR is targeted) and then do a merge in from `master` to get @colombod's changes that add a language parameter to the workspace object and clean up the various hacks I added around detecting the language.

File: FSharpWorkspaceShim/Library.fs (hunk 4 above)
Comment (brettfo) [reply FSharpWorkspaceShim/Library.fs]: Slightly shorter than `newlineificize`.

File: FSharpWorkspaceShim/FSharpWorkspaceShim.fsproj (hunk 3 above)
Comment (brettfo) [reply FSharpWorkspaceShim/FSharpWorkspaceShim.fsproj]: Thanks, I'll bring this in a subsequent PR.

File: WorkspaceServer/Servers/WorkspaceServerMultiplexer.cs [hunk 6]
```diff
@@ -0,0 +1,65 @@
...
+        public async Task<RunResult> Run(WorkspaceRequest request, Budget budget = null)
+        {
+            return await IsFSharpWorkspaceRequest(request.Workspace)
+                ? await _fsharpWorksapceServer.Run(request, budget)
+                : await _roslynWorkspaceServer.Run(request, budget);
+        }
+
+        private async Task<bool> IsFSharpWorkspaceRequest(Workspace workspace)
+        {
+            var package = await _packageFinder.Find<Package>(workspace.WorkspaceType);
```
colombod: there is langauge on the workspace, that shoudl tell you if you are looking at a csharp or fsharp request

WorkspaceServer/Servers/WorkspaceServerMultiplexer.cs
```diff
@@ -0,0 +1,65 @@
...
+using WorkspaceServer.Servers.Roslyn;
+using Package = WorkspaceServer.Packaging.Package;
+
+namespace WorkspaceServer.Servers
+{
+    public class WorkspaceServerMultiplexer : IWorkspaceServer
+    {
+        private IPackageFinder _packageFinder;
+        private IWorkspaceServer _roslynWorkspaceServer;
+        private IWorkspaceServer _fsharpWorksapceServer;
```
colombod: typo, also these two can be read only

PR #238, opened 2019-05-28
Blazor controller

rchande:


MLS.Agent.Tests/ApiViaHttpTests.cs
```diff
@@ -802,7 +802,40 @@ public async Task Can_serve_blazor_console_code_runner()
                 result.Should().Contain("Loading...");
             }
         }
-        
+
+        [Fact]
+        public async Task Can_serve_humanizer_code_runner()
```
jonsequitur: >blazor [](start = 41, length = 6)

webassembly

PR #246, opened 2019-05-30
Add --port option to dotnet try

akshita31:
1. In the Program.cs, based on if we are in tool mode or not we decide the url that will be used.
If we are in tool mode and the user has specified a port, we will use it otherwise we will launch on an available open port.
2. In the Startup.cs we use the `ServerFeatures` property to know the url at which the application is running and then launch the browser on the same when we are in the try mode.

### DIFF
--- a/MLS.Agent.Tests/AgentService.cs
+++ b/MLS.Agent.Tests/AgentService.cs
@@ -45,6 +45,8 @@ namespace MLS.Agent.Tests
 
         private TestServer CreateTestServer()
         {
+            //We need to set the feature collection to be not null,
+            //so that the ServerFeatures(like the urls to use) can be configured and used
             var featureCollection = new FeatureCollection();
             featureCollection.Set<IServerAddressesFeature>(new ServerAddressesFeature());
             return new TestServer(CreateWebHostBuilder(), featureCollection);
--- a/MLS.Agent.Tests/AgentService.cs
+++ b/MLS.Agent.Tests/AgentService.cs
@@ -70,7 +70,7 @@ namespace MLS.Agent.Tests
                           })
                           .UseTestEnvironment()
                           .UseStartup<Startup>()
-                          .WithConfiguredApplicationUrl(_options);
+                          .ConfigureUrlUsingPort(_options.Port);
 
             return builder;
         }
--- a/MLS.Agent.Tests/AgentService.cs
+++ b/MLS.Agent.Tests/AgentService.cs
@@ -68,7 +68,7 @@ namespace MLS.Agent.Tests
                           })
                           .UseTestEnvironment()
                           .UseStartup<Startup>()
-                          .WithConfigureApplicationUrl(_options);
+                          .WithConfiguredApplicationUrl(_options);
 
             return builder;
         }
--- a/MLS.Agent.Tests/AgentService.cs
+++ b/MLS.Agent.Tests/AgentService.cs
@@ -53,7 +53,6 @@ namespace MLS.Agent.Tests
         private IWebHostBuilder CreateWebHostBuilder()
         {
             var builder = new WebHostBuilder()
-                          .UseUrls("http://localhost:4242")
                           .ConfigureServices(c =>
                           {
                               if (_directoryAccessor != null)
@@ -68,7 +67,8 @@ namespace MLS.Agent.Tests
                               });
                           })
                           .UseTestEnvironment()
-                          .UseStartup<Startup>();
+                          .UseStartup<Startup>()
+                          .WithConfigureApplicationUrl(_options);
 
             return builder;
         }
--- a/MLS.Agent.Tests/AgentService.cs
+++ b/MLS.Agent.Tests/AgentService.cs
@@ -5,6 +5,8 @@ using System;
 using System.Net.Http;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Hosting.Server.Features;
+using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.TestHost;
 using Microsoft.Extensions.DependencyInjection;
 using MLS.Agent.CommandLine;
@@ -41,11 +43,17 @@ namespace MLS.Agent.Tests
 
         public void Dispose() => _disposables.Dispose();
 
-        private TestServer CreateTestServer() => new TestServer(CreateWebHostBuilder());
+        private TestServer CreateTestServer()
+        {
+            var featureCollection = new FeatureCollection();
+            featureCollection.Set<IServerAddressesFeature>(new ServerAddressesFeature());
+            return new TestServer(CreateWebHostBuilder(), featureCollection);
+        }
 
         private IWebHostBuilder CreateWebHostBuilder()
         {
             var builder = new WebHostBuilder()
+                          .UseUrls("http://localhost:4242")
                           .ConfigureServices(c =>
                           {
                               if (_directoryAccessor != null)
--- a/MLS.Agent.Tests/CommandLine/CommandLineParserTests.cs
+++ b/MLS.Agent.Tests/CommandLine/CommandLineParserTests.cs
@@ -224,10 +224,10 @@ namespace MLS.Agent.Tests.CommandLine
         [Fact]
         public void Negative_port_fails_the_parse()
         {
-            _parser.Parse("--port -1")
+            _parser.Parse("--port -8090")
                    .Errors
                    .Should()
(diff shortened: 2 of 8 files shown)


akshita31:
The approach followed above will not work in IIS. However the configuration that we have currently uses IIS express for hosted mode, hence this code path is not called in that case.

akshita31:
image

MLS.Agent.Tests/CommandLine/CommandLineParserTests.cs
```diff
@@ -214,6 +214,20 @@ public async Task AiKey_defaults_to_null()
...
+        {
+            await _parser.InvokeAsync("--port 6000", _console);
+            _start_options.Port.Should().Be(6000);
+        }
+
+        [Fact]
+        public async Task Takes_5000_as_the_default_port()
+        {
+            await _parser.InvokeAsync("", _console);
+            _start_options.Port.Should().Be(5000);
```
jonsequitur: We might want to consider changing this from the default Kestrel ports. 

Also keep in mind we also listen on 5001 by default for https.

MLS.Agent/CommandLine/CommandLineParser.cs
```diff
@@ -181,6 +181,11 @@ RootCommand StartInTryMode()
                                       "Enable verbose logging to the console",
                                       new Argument<bool>()));
 
+                command.AddOption(new Option(
+                                        "--port",
+                                        "Specify the port to launch the dotnet try process",
```
jonsequitur: Maybe something a little more casual like: "Specify the port for dotnet try to listen on"?

MLS.Agent/CommandLine/CommandLineParser.cs
```diff
@@ -181,6 +181,11 @@ RootCommand StartInTryMode()
...
+                                        new Argument<int>()));
```
jonsequitur: A default value can be specified here. This is more helpful because it can be displayed in help and in the parse diagram.

MLS.Agent/Startup.cs
```diff
@@ -182,8 +183,9 @@ public void ConfigureServices(IServiceCollection services)
 
                 if (StartupOptions.Mode == StartupMode.Try)
                 {
+                    var uri = new Uri(app.ServerFeatures.Get<IServerAddressesFeature>().Addresses.First());
```
jonsequitur: Can there be more than one?

MLS.Agent.Tests/AgentService.cs
```diff
@@ -41,11 +43,17 @@ public AgentService(StartupOptions options = null, IDirectoryAccessor directoryA
 
         public void Dispose() => _disposables.Dispose();
 
-        private TestServer CreateTestServer() => new TestServer(CreateWebHostBuilder());
+        private TestServer CreateTestServer()
+        {
+            var featureCollection = new FeatureCollection();
+            featureCollection.Set<IServerAddressesFeature>(new ServerAddressesFeature());
+            return new TestServer(CreateWebHostBuilder(), featureCollection);
```
jonsequitur: Ideally this should not be a special set up that happens in test one way and in the application's startup code in another way.

MLS.Agent/Program.cs
```diff
@@ -139,9 +143,16 @@ public static IWebHost ConstructWebHost(StartupOptions options)
                           })
                           .UseEnvironment(options.EnvironmentName)
                           .UseStartup<Startup>()
+                          .UseUrls(uri.ToString())
                           .Build();
 
             return webHost;
         }
+
+        private static bool IsInToolMode()
```
jonsequitur: I think this should be called something like `IsLaunchedForDevelopment` or something like that. Basically this will be true when someone launches `dotnet-try` from source in Visual Studio, Visual Studio Code, or `dotnet run`.

MLS.Agent/Startup.cs
akshita31: Since we are setting up in the Program.cs to usea single url, there will be only one here - https://github.com/dotnet/try/pull/246/files#diff-adcfc4b0634e7250308be781cbe69f63R146

MLS.Agent.Tests/AgentService.cs
akshita31: https://github.com/dotnet/try/pull/246/files#diff-cd33a75e1de2941d5f8cbf5c1494709dL63.

The setup is using the same extension method as the main Program to configure the application url

MLS.Agent/Startup.cs
akshita31: https://github.com/dotnet/try/pull/246/files#diff-7e4af26b8c2fa96387cebaf1ae6d7716R14

MLS.Agent.Tests/AgentService.cs
akshita31: https://github.com/dotnet/try/pull/246/files#diff-7e4af26b8c2fa96387cebaf1ae6d7716R14

akshita31: For more context, since we are reading the url at which the application is being launched, using the ServerFeatures here: https://github.com/dotnet/try/pull/246/files#diff-ca9274b169fb9c6f45c3159115b91007R186, we need to set the feature collection in the TestServer to not to be null. https://github.com/aspnet/Hosting/pull/969

MLS.Agent/WebHostBuilderExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        private static bool IsLaunchedForDevelopment()
+        {
+            var processName = System.Diagnostics.Process.GetCurrentProcess().ProcessName;
+            return processName == "dotnet" || processName == "dotnet.exe";
+        }
+
+        private static int GetFreePort()
+        {
+            TcpListener l = new TcpListener(IPAddress.Loopback, 0);
```
rchande: Is the second argument the port it starts searching from? Might be good to pick some value like `7000` or something so that the browser windows we open point to similar ports.

rchande: (I'm not sure what the availability/spacing is of open ports starting at 0 on the various oses we support)

MLS.Agent/CommandLine/StartupOptions.cs
```diff
@@ -42,7 +42,8 @@ public static StartupOptions FromCommandLine(string commandLine)
             bool enablePreviewFeatures = false,
             string package = null,
             string packageVersion = null,
-            ParseResult parseResult = null)
+            ParseResult parseResult = null,
+            string port = null)
```
jonsequitur: >string [](start = 12, length = 6)

This should be int.

akshita31: If it is int then we will have to specify some value. What should that be ?

MLS.Agent/WebHostBuilderExtensions.cs
akshita31: That is the "listening socket". I tried to specify 700 in place of 0 there and it looks like if we do that it tries to open that particular port only. Specifying 0 I think implies that look for the next available free port.

MLS.Agent/CommandLine/StartupOptions.cs
akshita31: Also, I am checking if this value or not to determine if we need to use this port or give a random port. I can make this as int and set the default value to 0 and then check that if options.Port is 0 then we should start on a random port. How does that sound?

jonsequitur: If you want to distinguish between whether it was specified or not, you could use `int?`.

akshita31: Done. Thanks!

MLS.Agent/CommandLine/CommandLineParser.cs
```diff
@@ -181,6 +181,11 @@ RootCommand StartInTryMode()
                                       "Enable verbose logging to the console",
                                       new Argument<bool>()));
 
+                command.AddOption(new Option(
+                                        "--port",
+                                        "Specify the port for dotnet try to listen on",
+                                        new Argument<int>()));
```
colombod: negative numbers are not valid

rchande: Type should probably be `ushort`, since that matches the sign/max value requirements.

MLS.Agent/WebHostBuilderExtensions.cs
rchande: Cool, makes sense

akshita31:
> Is there a way for us to have an ApiViaHttpTest for the auto port selection?

I have added a unit test around the part that determines the urls

PR #252, opened 2019-06-03
check for null HtmlNode and report error

brettfo:
Some tests have been flaky because `HtmlDocument.DocumentNode` was null which indicates an issue parsing the HTML.  In an attempt to discover the issue I've added a wrapper that reports the parsed HTML if `DocumentNode` was null.

Marking as work-in-progress because I expect to reset the tests repeatedly.


brettfo:
The `null` value seems to come from `.SelectSingleNode()`.  One failing test which was looking for `"//pre/code"` had the HTML contents of:

``` html
<div class="notification is-danger">
  <svg ...>
  </svg>
  Package named "console" not found.
</div>
```

So the missing packages appears to be the root cause of most test flakiness.

jonsequitur:
I think this is a concurrency issue on creation of these package assets on-disk during test runs.

PR #254, opened 2019-06-04
enable language as property of workspaces

colombod:
language detected in markdown fences is the pushed to html and the js api makes it flow as part of the protocol back into agent.

dotnet try verify detects incompatibility between project and language

### DIFF
--- a/MLS.Agent.Tests/CommandLine/VerifyCommandTests.cs
+++ b/MLS.Agent.Tests/CommandLine/VerifyCommandTests.cs
@@ -146,7 +146,7 @@ This is some sample code:
                     ("Program.cs", CompilingProgramCs),
                     ("support.fs", "let a = 0"),
                     ("doc.md", @"
-```fs --source-file support.fs
+```fs --source-file support.fs --project some.csproj
 ```
 ")
                 }.CreateFiles();
--- a/MLS.Agent.Tests/CommandLine/VerifyCommandTests.cs
+++ b/MLS.Agent.Tests/CommandLine/VerifyCommandTests.cs
@@ -136,6 +136,41 @@ This is some sample code:
             }
 
             [Fact]
+            public async Task Fails_if_language_is_not_compatible_with_backing_project()
+            {
+                var root = Create.EmptyWorkspace(isRebuildablePackage: true).Directory;
+
+                var directoryAccessor = new InMemoryDirectoryAccessor(root, root)
+                {
+                    ("some.csproj", CsprojContents),
+                    ("Program.cs", CompilingProgramCs),
+                    ("support.fs", "let a = 0"),
+                    ("doc.md", @"
+```fs --source-file support.fs
+```
+")
+                }.CreateFiles();
+
+                var console = new TestConsole();
+
+                await VerifyCommand.Do(
+                    new VerifyOptions(root),
+                    console,
+                    () => directoryAccessor,
+                    PackageRegistry.CreateForTryMode(root));
+
+                _output.WriteLine(console.Out.ToString());
+
+                console.Out
+                    .ToString()
+                    .EnforceLF()
+                    .Trim()
+                    .Should()
+                    .Match(
+                        $"*Build failed as project {root}{Path.DirectorySeparatorChar}some.csproj is not compatible with language fsharp*".EnforceLF());
+            }
+
+            [Fact]
             public async Task When_non_editable_code_blocks_do_not_contain_errors_then_validation_succeeds()
             {
                 var root = Create.EmptyWorkspace(isRebuildablePackage: true).Directory;
--- a/MLS.Agent.Tests/Markdown/CodeBlockAnnotationExtensionTests.cs
+++ b/MLS.Agent.Tests/Markdown/CodeBlockAnnotationExtensionTests.cs
@@ -67,14 +67,14 @@ namespace BasicConsoleApp
             var document =
 $@"```{language} --source-file Program.cs
 ```";
-            string html = (await pipeline.RenderHtmlAsync(document)).EnforceLF();
+            var html = (await pipeline.RenderHtmlAsync(document)).EnforceLF();
             html.Should().Contain(fileContent.HtmlEncode().ToString());
         }
 
         [Fact]
-        public async Task Does_not_insert_code_when_specified_language_is_not_csharp()
+        public async Task Does_not_insert_code_when_specified_language_is_not_supported()
         {
-            string expectedValue =
+            var expectedValue =
 @"<pre><code class=""language-js"">console.log(&quot;Hello World&quot;);
 </code></pre>
 ".EnforceLF();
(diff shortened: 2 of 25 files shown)


MLS.Agent.Tests/CommandLine/VerifyCommandTests.cs
```diff
@@ -135,6 +135,41 @@ public async Task Files_are_listed()
...
+                var directoryAccessor = new InMemoryDirectoryAccessor(root, root)
+                {
+                    ("some.csproj", CsprojContents),
+                    ("Program.cs", CompilingProgramCs),
+                    ("support.fs", "let a = 0"),
+                    ("doc.md", @"
+```fs --source-file support.fs
+```
+")
+                }.CreateFiles();
```
jonsequitur: Can this test do without creating the files on disk or is the `verify` command still dependent on real project files?

MLS.Agent.Tests/Markdown/CodeBlockAnnotationExtensionTests.cs
```diff
@@ -90,27 +90,34 @@ public async Task Does_not_insert_code_when_specified_language_is_not_csharp()
...
+        [InlineData("c#", "language-c#")]
+        [InlineData("fs", "language-fs")]
+        [InlineData("fsharp", "language-fsharp")]
+        [InlineData("f#", "language-f#")]
+        public async Task Does_not_insert_code_when_supported_language_is_specified_but_no_additional_options(string fenceLanguage, string expectedClass)
         {
-            string expectedValue =
-@"<pre><code class=""language-cs"">Console.WriteLine(&quot;Hello World&quot;);
+            var expectedValue =
+$@"<pre><code class=""{expectedClass}"">Console.WriteLine(&quot;Hello World&quot;);
```
jonsequitur: Should we not normalize these so that the output class is always the same for the various   inputs from the fence?

MLS.Agent/CommandLine/VerifyCommand.cs
```diff
@@ -257,5 +271,28 @@ void SetOk()
                 }
             }
         }
+
+        private static bool ProjectIsCompatibleWithLanguage(string projectOrPackageName, string language)
+        {
+            var extenstion = Path.GetExtension(projectOrPackageName)?.ToLowerInvariant();
```
jonsequitur: Spelling: "extension"

MLS.Agent/Markdown/AnnotatedCodeBlockExtensions.cs
```diff
@@ -30,5 +30,10 @@ public static string ProjectOrPackageName(this AnnotatedCodeBlock block)
                 (block.Annotations as LocalCodeBlockAnnotations)?.Project?.FullName ??
                 block.Annotations?.Package;
         }
+        public static string Language(this AnnotatedCodeBlock block)
+        {
+            return
+                (block.Annotations as LocalCodeBlockAnnotations)?.NormalizedLanguage;
```
jonsequitur: Why is this constrained to `LocalCodeBlockAnnotations`?

Microsoft.DotNet.Try.Client/src/IState.ts
```diff
@@ -140,6 +140,7 @@ export interface IWorkspaceState
 
 export interface IWorkspace {
     workspaceType: string;
+    langauge?: string;
```
jonsequitur: Spelling.

Microsoft.DotNet.Try.Markdown/CodeBlockAnnotations.cs
```diff
@@ -49,6 +50,7 @@ public class CodeBlockAnnotations
         public bool Editable { get; }
         public bool Hidden { get; }
         public string Language { get; set; }
+        public string NormalizedLanguage { get; set; }
```
jonsequitur: Can this be read-only?

Microsoft.DotNet.Try.Markdown/CodeFenceAnnotationsParser.cs
```diff
@@ -82,37 +82,49 @@ public class CodeFenceAnnotationsParser
...
-            if (defaultAnnotations?.PackageVersion is string defaultPackageVersion)
+            if (_defaultAnnotations?.PackageVersion is string defaultPackageVersion)
             {
                 packageVersionOption.Argument.SetDefaultValue(defaultPackageVersion);
             }
 
+            var languageCommands = new[]
+            {
+                CreateCsharpCommand(configureCsharpCommand, packageOption, packageVersionOption),
+                CreateFsharpCommand(configureCsharpCommand, packageOption, packageVersionOption)
```
jonsequitur: This shouldn't pass `configureCsharpCommand` or else the parameter should be renamed if the configurations are intended to apply to all subcommands.

Microsoft.DotNet.Try.Markdown/CodeFenceAnnotationsParser.cs
```diff
@@ -121,8 +133,36 @@ private Parser CreateOptionsParser(Action<Command> configureCsharpCommand = null
...
             csharp.AddAlias("cs");
             csharp.AddAlias("c#");
+            return csharp;
+        }
 
-            return new Parser(new RootCommand { csharp });
+        private static Command CreateFsharpCommand(Action<Command> configureCsharpCommand, Option packageOption,
+            Option packageVersionOption)
+        {
+            var fsharp = new Command("fsharp")
```
jonsequitur: Please refactor to avoid duplication. Anything that's shared between these commands would be better applied through a common method.

Microsoft.DotNet.Try.Protocol.ClientApi/Project.cs
```diff
@@ -12,12 +12,16 @@ namespace Microsoft.DotNet.Try.Protocol.ClientApi
     [JsonConverter(typeof(ProjectJsonConverter))]
     public class Project : FeatureContainer
     {
+        private const string DefaultLanguage = "csharp";
```
jonsequitur: Is there an approach where we don't have the concept of a default language?

MLS.Agent/CommandLine/VerifyCommand.cs
```diff
@@ -257,5 +271,28 @@ void SetOk()
                 }
             }
         }
+
+        private static bool ProjectIsCompatibleWithLanguage(string projectOrPackageName, string language)
```
rchande: If this string really is a project or package name you shouldn't assume it's a file path...

MLS.Agent/CommandLine/VerifyCommand.cs
```diff
@@ -257,5 +271,28 @@ void SetOk()
...
+        {
+            var extenstion = Path.GetExtension(projectOrPackageName)?.ToLowerInvariant();
+            var supported = true;
+            if (!string.IsNullOrWhiteSpace(extenstion))
+            {
+                switch (extenstion)
+                {
+                    case ".csproj":
+                        supported = StringComparer.OrdinalIgnoreCase.Compare(language, "csharp") == 0;
```
rchande: Worth creating a language enum?

Microsoft.DotNet.Try.Protocol.ClientApi/Project.cs
rchande: +1

MLS.Agent.Tests/Markdown/CodeBlockAnnotationExtensionTests.cs
colombod: sounds a good idea.

Microsoft.DotNet.Try.Protocol.ClientApi/Project.cs
colombod: is for backward compatibility

MLS.Agent/CommandLine/VerifyCommand.cs
jonsequitur: Maybe we want a custom type for this similar to the approach we discussed for `FileInfoOrUri`. Where did that land?

MLS.Agent/CommandLine/VerifyCommand.cs
jonsequitur: I think eventually we may want this to be open-ended and pluggable at which point the enum would wear out its welcome. Also, the enum doesn't help with authoring the actual Markdown.

MLS.Agent.Tests/CommandLine/VerifyCommandTests.cs
colombod: it si still looking for real file in verify

Microsoft.DotNet.Try.Markdown/CodeBlockAnnotations.cs
colombod: making class immutable then

MLS.Agent/CommandLine/VerifyCommand.cs
rchande: "PackageSource"? I think...

rchande: Just move that type to the right layer and rename I guess

PR #258, opened 2019-06-06
Make wasm package ingestion actually work

rchande:


WorkspaceServer/Packaging/WebAssemblyAssetFinder.cs
```diff
@@ -37,12 +38,18 @@ async Task<TPackage> IPackageFinder.Find<TPackage>(PackageDescriptor descriptor)
...
             await tool.Prepare();
-            var wasmAsset = await tool.LocateWasmAsset();
-            if (wasmAsset != null)
+            var buildAsset = await tool.LocateProjectAsset();
+            if (buildAsset != null)
             {
-                var package = new Package2(descriptor.Name, new FileSystemDirectoryAccessor(wasmAsset.DirectoryAccessor.GetFullyQualifiedRoot().Parent));
-                package.Add(wasmAsset);
-                return package;
+                var package = new Package2(descriptor.Name, new FileSystemDirectoryAccessor(buildAsset.DirectoryAccessor.GetFullyQualifiedRoot().Parent));
```
jonsequitur: >new FileSystemDirectoryAccessor(buildAsset.DirectoryAccessor.GetFullyQualifiedRoot().Parent) [](start = 60, length = 92)

Moving to a parent directory from the asset seems like a can of worms that I feel breaks the abstraction.

We have a method, DirectoryAccessor.GetDirectoryAccessorForRelativePath, which I feel we should be using here. WebAssemblyAssetFinder should take an IDirectoryAccessor in its constructor and build one accessor from another so when we test it via the in-memory implementation, it won't break out into a file-based implementation.

WorkspaceServer/WorkspaceFeatures/PackageTool.cs
```diff
@@ -28,11 +28,11 @@ public PackageTool(string name, DirectoryInfo workingDirectory)
 
         public string Name { get; }
 
-        public async Task<DirectoryInfo> LocateBuildAsset()
+        public async Task<ProjectAsset> LocateProjectAsset()
         {
             var result = await CommandLine.Execute(_path.Value, MLS.PackageTool.PackageToolConstants.LocateBuildAsset, _workingDirectory);
             var projectDirectory = new DirectoryInfo(string.Join("", result.Output));
-            return projectDirectory;
+            return new ProjectAsset(new FileSystemDirectoryAccessor(projectDirectory));
```
jonsequitur: >new FileSystemDirectoryAccessor(projectDirectory) [](start = 36, length = 49)

Similar issue. We should inject IDirectoryAccessor and move in the direction of in-memory testing for these things.

WorkspaceServer/Packaging/WebAssemblyAssetFinder.cs
```diff
@@ -37,12 +38,18 @@ async Task<TPackage> IPackageFinder.Find<TPackage>(PackageDescriptor descriptor)
         protected async Task<IPackage> CreatePackage(PackageDescriptor descriptor, PackageTool tool)
         {
             await tool.Prepare();
-            var wasmAsset = await tool.LocateWasmAsset();
-            if (wasmAsset != null)
+            var buildAsset = await tool.LocateProjectAsset();
```
akshita31: >buildAsset [](start = 16, length = 10)

Nit: We can call it projectAsset

WorkspaceServer/WorkspaceFeatures/PackageTool.cs
```diff
@@ -28,11 +28,11 @@ public PackageTool(string name, DirectoryInfo workingDirectory)
 
         public string Name { get; }
 
-        public async Task<DirectoryInfo> LocateBuildAsset()
+        public async Task<ProjectAsset> LocateProjectAsset()
         {
             var result = await CommandLine.Execute(_path.Value, MLS.PackageTool.PackageToolConstants.LocateBuildAsset, _workingDirectory);
```
akshita31: >LocateBuildAsset [](start = 101, length = 16)

Does this also need to change to "LocateProjectAsset" ?

PR #261, opened 2019-06-07
jupyter protocol

colombod:
complete message implementation


Microsoft.DotNet.Try.Jupyter/IMessageBuilder.cs
```diff
@@ -1,21 +1,41 @@
...
+using System;
 using System.Collections.Generic;
 using Microsoft.DotNet.Try.Jupyter.Protocol;
 
 namespace Microsoft.DotNet.Try.Jupyter
 {
     public interface IMessageBuilder
     {
-        Header CreateHeader(string messageType, string session);
+        Message CreateMessage(JupyterMessageContent content, Header parentHeader, List<byte[]> identifiers = null);
```
jonsequitur: Does this need to be a list? `ReadOnlySpan<byte>` maybe?

MLS.Agent.Tests/ApiViaHttpTestsBase.cs
```diff
@@ -63,7 +63,7 @@ protected ApiViaHttpTestsBase(ITestOutputHelper output)
             StartupOptions options = null)
         {
             HttpResponseMessage response;
-            using (var agent = new AgentService(options ?? StartupOptions.FromCommandLine("hosted")))
+            using (var agent = new AgentService(options ?? StartupOptionsFactory.CreateFromCommandLine("hosted")))
```
jonsequitur: Why introduce a factor class for this? I think this makes it less discoverable.

MLS.Agent/Program.cs
```diff
@@ -27,11 +26,11 @@ namespace MLS.Agent
 {
     public class Program
     {
-        private static readonly ServiceCollection _serviceCollection = new ServiceCollection();
+        private static readonly ServiceCollection ServiceCollection = new ServiceCollection();
```
jonsequitur: The previous name is the normal format for fields. Hungarian-ish, but we should be consistent.

Microsoft.DotNet.Try.Jupyter/JupyterRequestContextHandler.cs
```diff
@@ -30,155 +35,190 @@ public JupyterRequestContextHandler(PackageRegistry packageRegistry)
...
-                                delivery.Command.Request.Header);
-                            ioPubChannel.Send(stream);
-                        }
-                    }
-
-                    if (!executeRequest.Silent && resultSucceeded)
+            // executeResult data
+            var executeResultData = new ExecuteResult
+            {
+                Data = new JObject
```
jonsequitur: Can we use a more specific type here? Mixing our own message exchange types with `NewtonSoft.Json` types is awkward.

MLS.Agent.Tests/ApiViaHttpTestsBase.cs
colombod: this FromCommandLine functionality is only used in tests. What would be a better name for that?

MLS.Agent/Program.cs
colombod: static and const tend to be with that naming (it might be something inherited from resharper settings)

Microsoft.DotNet.Try.Jupyter/JupyterRequestContextHandler.cs
colombod: Data is a "python dictionary" can try to use dictionary of string,object

MLS.Agent/Program.cs
jonsequitur: I think it's only when not `private`.

MLS.Agent.Tests/ApiViaHttpTestsBase.cs
jonsequitur: I see. Personally, I'd still leave it in the product code so people stumble onto it more easily.

Microsoft.DotNet.Try.Jupyter/JupyterRequestContextHandler.cs
```diff
@@ -30,155 +35,190 @@ public JupyterRequestContextHandler(PackageRegistry packageRegistry)
...
+                    {"text/html", output},
+                    {"text/plain", output}
+                },
+                Transient = transient,
+                ExecutionCount = _executionCount
+            };
+
+
+            var resultSucceeded = result.Succeeded &&
+                                  result.Exception == null;
```
jonsequitur: Are we (somewhere) constraining this so that `Succeeded` must be false when there's an `Exception`? The possibility that we would have `Succeeded == true` and an exception looks wrong.

Microsoft.DotNet.Try.Jupyter/JupyterRequestContextHandler.cs
jonsequitur: Or maybe create a custom type that just inherits dictionary but is a bit more of a signpost to where this abstraction comes from?

Microsoft.DotNet.Try.Jupyter/IMessageBuilder.cs
colombod: is a list of byte []

jonsequitur: Meaning it needs to be mutable?

Microsoft.DotNet.Try.Jupyter.Tests/HistorySerializationTests.cs
```diff
@@ -0,0 +1,77 @@
...
+namespace Microsoft.DotNet.Try.Jupyter.Tests
+{
+    public class HistorySerializationTests
+    {
+        [Fact]
+        public void Can_serialize_input_and_output_element()
+        {
+            var element = new InputOutputHistoryElement(0, 0, "input value", "output result");
+            var serialized = JsonConvert.SerializeObject(element);
+            serialized.Should().Be(@"[0,0,[""input value"",""output result""]]");
```
jonsequitur: Verifying round-trip serializability for these tests will be a bit lower maintenance. For example, reordering fields will impact the JSON string but not the correctness.

Microsoft.DotNet.Try.Jupyter/Protocol/InputHistoryElementConverter.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            var source = serializer.Deserialize<JArray>(reader);
+            if (source.Count != 3)
+            {
+                throw new FormatException("the input must be a tuple with 3 elements");
+            }
+
+            var session = source[0].Value<int>();
+            var lineNumber = source[1].Value<int>();
+            var input = source[2].Value<string>();
```
jonsequitur: Why is it ordering-dependent?

colombod: because tuple serialization from python will make arrays with values so order matter (is not a record)

Microsoft.DotNet.Try.Jupyter.Tests/HistorySerializationTests.cs
colombod: this this message type reordering will break things. so it must be respected

Microsoft.DotNet.Try.Jupyter/JupyterRequestContextHandler.cs
colombod: at [RoslynWorkspaceServer.cs](https://github.com/colombod/try/blob/e22606aff72146dfa7a41113d83523e049350a52/WorkspaceServer/Servers/Roslyn/RoslynWorkspaceServer.cs#L338)

we can have 

` var runResult = new RunResult(
                succeeded: true,
                output: output.StdOut,
                exception: exceptionMessage,
                diagnostics: diagnostics,
                requestId: requestId);`

with `exceptionMessage` not null

Issue #262, opened 2019-06-07
Images don't work


jimbobbennett:
#### Describe the bug
Create markdown referencing a local image file.
The image is not found.

#### Did this error occur while using `dotnet try` or online?
- [X] `dotnet-try`
- [ ] online

#### What kind of error was it?
- [X] User Interface (UI):  For example the output never displayed 
- [ ] Service Error: For example "The service is temporarily unavailable. We are working on it"
- [ ] Other:

#### Screenshots
<img width="1052" alt="2019-06-07_17-46-38" src="https://user-images.githubusercontent.com/1710385/59120042-4cc4da80-894c-11e9-9768-5886684f32e1.png">

#### Please complete the following:
 - OS
    - [ ] Windows 10
    - [X] macOS
    - [ ] Linux (Please specify distro)
    - [ ] iOS
    - [ ] Android
 - Browser 
    - [ ] Chrome
    - [ ] Edge
    - [X] Safari

jimbobbennett:
@akshita31 - will your changes in #268 also allow css and js files to be referenced?

akshita31:
@jimbobbennett Yes javascript and css will also work as we are using the Static file handler from Asp.net. I have added a test in the PR that checks if a javascript file can be served.

PR #268, opened 2019-06-10
Making Images work

akshita31:
Fixed: https://github.com/dotnet/try/issues/262

Added the RootDirectoryPath as one of the paths to serve static files from. 
Also reordered so that the DocumentationController gets precedence to render the markdown files


WorkspaceServer/IDirectoryAccessor.cs
```diff
@@ -19,6 +19,8 @@ public interface IDirectoryAccessor
 
         string ReadAllText(RelativeFilePath path);
 
+        byte[] ReadAllBytes(RelativeFilePath path);
```
jonsequitur: This doesn't need to be on the interface. It can be an extension method.

WorkspaceServer.Tests/TestProjects/SampleConsole/a.html
```diff
@@ -0,0 +1 @@
+<p>This file should be served</p>
```
jonsequitur: It would be better to move the test file to the project that actually depends on it. It's unrelated to the scenario for which the `SampleConsole` directory was created, which is to test the `IDirectoryAccessor`.

MLS.Agent.Tests/ApiViaHttpTests.cs
```diff
@@ -1018,6 +1018,22 @@ public async Task Scaffolding_HTML_trydotnet_js_autoEnable_useWasmRunner_is_true
...
+        {
+            var options = new StartupOptions(dir: TestAssets.SampleConsole);
+
+            using (var agent = new AgentService(options: options))
+            {
+                var response = await agent.GetAsync(@"/a.html");
+
+                response.Should().BeSuccessful();
+                var html = await response.Content.ReadAsStringAsync();
+                html.Should().Be("<p>This file should be served</p>");
```
jonsequitur: Since the Markdown parser will pass HTML through, this test doesn't actually ensure that the static file was served. If the Markdown route matches this URL, you'll see the same result.

MLS.Agent.Tests/MLS.Agent.Tests.csproj
```diff
@@ -10,6 +10,12 @@
     <EmbeddedResource Remove="TestResults\**" />
     <None Remove="TestResults\**" />
   </ItemGroup>
+
+  <ItemGroup>
+    <None Include="TestAssets\**">
```
rchande: Is there only one file in TestAssets? In my opinion, the test that uses this might be more readable if it used our DisposableDirectory abstraction to create a temp directory with a single file and verified that serving static files worked that way. Otherwise, we've added yet another file that gets copied around by our build...

MLS.Agent.Tests/MLS.Agent.Tests.csproj
```diff
@@ -10,6 +10,12 @@
...
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
```
rchande: If we decide to go this way, might as well set this to "IfNewer"

MLS.Agent.Tests/TestAssets/a.js
```diff
@@ -0,0 +1 @@
+alert('This is an alert from javascript');
```
rchande: Copyright header? (I have probably forgotten some myself)

MLS.Agent.Tests/MLS.Agent.Tests.csproj
akshita31: Thanks! I have updated to use the "DisposableDirectory"

PR #270, opened 2019-06-11
Use the project asset only if the command line succeeded

akshita31:
We should check if the command line execution succeeded before we try to use the output from the results

### DIFF
--- a/WorkspaceServer/WorkspaceFeatures/PackageTool.cs
+++ b/WorkspaceServer/WorkspaceFeatures/PackageTool.cs
@@ -4,6 +4,7 @@
 using MLS.Agent.Tools;
 using System.IO;
 using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices.ComTypes;
 using System.Threading.Tasks;
 using WorkspaceServer.Packaging;
 
@@ -42,7 +43,7 @@ namespace WorkspaceServer.WorkspaceFeatures
             {
                 return new ProjectAsset(new FileSystemDirectoryAccessor(projectDirectory));
             }
-            
+
             return null;
         }
 
@@ -51,16 +52,13 @@ namespace WorkspaceServer.WorkspaceFeatures
             if (Exists())
             {
                 var result = await CommandLine.Execute(FilePath, command, DirectoryAccessor.GetFullyQualifiedRoot());
-                if (result.ExitCode == 0)
+                var path = string.Join("", result.Output);
+                if (result.ExitCode == 0 && !string.IsNullOrEmpty(path))
                 {
-                    var path = string.Join("", result.Output);
-                    if (!string.IsNullOrEmpty(path))
+                    var projectDirectory = new DirectoryInfo(string.Join("", result.Output));
+                    if (projectDirectory.Exists)
                     {
-                        var projectDirectory = new DirectoryInfo(string.Join("", result.Output));
-                        if (projectDirectory.Exists)
-                        {
-                            return projectDirectory;
-                        }
+                        return projectDirectory;
                     }
                 }
             }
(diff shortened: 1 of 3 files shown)


WorkspaceServer/WorkspaceFeatures/PackageTool.cs
```diff
@@ -31,21 +31,29 @@ public PackageTool(string name, IDirectoryAccessor directoryAccessor)
...
+            if (result.ExitCode == 0)
+            {
+                var projectDirectory = new DirectoryInfo(string.Join("", result.Output));
+                return new ProjectAsset(new FileSystemDirectoryAccessor(projectDirectory));
+            }
+
+            return null;
         }
 
         public async Task<WebAssemblyAsset> LocateWasmAsset()
```
rchande: I would also assert that Exists is true.

WorkspaceServer/WorkspaceFeatures/PackageTool.cs
```diff
@@ -31,21 +31,29 @@ public PackageTool(string name, IDirectoryAccessor directoryAccessor)
...
-                return null;
+                var projectDirectory = new DirectoryInfo(string.Join("", result.Output));
+
+                if (projectDirectory.Exists)
+                {
+                    return new WebAssemblyAsset(new FileSystemDirectoryAccessor(projectDirectory));
+                }
             }
 
-            return new WebAssemblyAsset(new FileSystemDirectoryAccessor(projectDirectory));
```
rchande: Does the calling side handle returning null? Would it be better to throw here so that we can throw an informative exception (instead of NRE in the caller)?

WorkspaceServer/WorkspaceFeatures/PackageTool.cs
```diff
@@ -34,7 +34,10 @@ public async Task<ProjectAsset> LocateProjectAsset()
             if (result.ExitCode == 0)
             {
                 var projectDirectory = new DirectoryInfo(string.Join("", result.Output));
-                return new ProjectAsset(new FileSystemDirectoryAccessor(projectDirectory));
+                if (projectDirectory.Exists)
```
rchande: I actually meant that we should asset that PackageTool.Exists is true

WorkspaceServer/WorkspaceFeatures/PackageTool.cs
akshita31: The caller here is not sure that the asset certainly exists and calls this function and if the function returns null it handles that https://github.com/dotnet/try/blob/3fedb8fa682242fa466ed2b826fb8df934a4e465/WorkspaceServer/Packaging/WebAssemblyAssetFinder.cs#L41

If we throw an exception here, should the caller catch the exception ?

jonsequitur: Whose error is it if the asset doesn't exist?

rchande: Made this a little clearer with https://github.com/dotnet/try/pull/273

WorkspaceServer/WorkspaceFeatures/PackageTool.cs
```diff
@@ -30,22 +30,39 @@ public PackageTool(string name, IDirectoryAccessor directoryAccessor)
...
         {
-            var result = await CommandLine.Execute(_path.Value, MLS.PackageTool.PackageToolConstants.LocateProjectAsset, DirectoryAccessor.GetFullyQualifiedRoot());
-            var projectDirectory = new DirectoryInfo(string.Join("", result.Output));
-            return new ProjectAsset(new FileSystemDirectoryAccessor(projectDirectory));
+            if (Exists)
+            {
+                var result = await CommandLine.Execute(_path.Value, MLS.PackageTool.PackageToolConstants.LocateProjectAsset, DirectoryAccessor.GetFullyQualifiedRoot());
+                if (result.ExitCode == 0)
+                {
+                    var projectDirectory = new DirectoryInfo(string.Join("", result.Output));
```
brettfo: Given the failure modes I've seen in the past it would be nice if this checked for an empty string, otherwise internally `Path.GetFullPath()` will throw and it's not really helpful.  E.g.,

``` c#
var path = string.Join("", result.Output);
if (string.IsNullOrEmpty(path))
{
    // throw, return null, something else
}
var projectDirectory = new DirectoryInfo(path);
```

PR #275, opened 2019-06-11
Remove the "template" word from the template

akshita31:


Microsoft.DotNet.Try.ProjectTemplate/Tutorial/content/.template.config/template.json
```diff
@@ -13,7 +12,7 @@
       },
     "identity": "Microsoft.DotNet.Try.ProjectTemplate.Tutorial",
     "defaultName": "Microsoft.DotNet.Try.ProjectTemplate.Tutorial",
-    "name": "Interactive documentation template",
+    "name": ".NET Interactive documentation",
```
jonsequitur: I think we can do without the ".NET" part, since when someone types `dotnet new list` they already have that context.

akshita31: So just "Interactive documentation" right ?

PR #276, opened 2019-06-11
Don't create packages in package registry

rchande:


WorkspaceServer.Tests/Default.cs
```diff
@@ -14,12 +14,19 @@ public class Default
     {
         public static AsyncLazy<PackageRegistry> PackageFinder { get; } = new AsyncLazy<PackageRegistry>(async () =>
```
jonsequitur: The method name is misleading.

PR #291, opened 2019-06-19
add microsoft as author

akshita31:


eng/targets/Settings.props
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
-    <Authors>dotnet organization</Authors>
+    <Authors>Microsoft Corporation</Authors>
```
brettfo:
```suggestion
    <Authors>Microsoft</Authors>
```

The other `dotnet/*` repos just use 'Microsoft'.

PR #295, opened 2019-06-21
domain events and repl model

colombod:


WorkspaceServer.Tests/Kernel/CSharpReplTests.cs
```diff
@@ -0,0 +1,139 @@
...
+            await repl.SendAsync(new SubmitCode("var a ="));
+
+            await repl.SendAsync(new SubmitCode("12;"));
+
+            KernelEvents.Should()
+                .NotContain(e => e is ValueProduced);
+
+            KernelEvents
+                .Should()
+                .Contain(e => e is CodeSubmissionEvaluated);
```
akshita31: Since in the test above this we have "CodeSubmissionEvaluationFailed", should we have here "CodeSubmissionEvaluationSucceded" ?

WorkspaceServer.Tests/Kernel/CSharpReplTests.cs
```diff
@@ -0,0 +1,139 @@
...
+        }
+
+        [Fact]
+        public async Task it_does_not_return_a_result_for_a_statement()
+        {
+            var repl = await CreateKernelAsync();
+
+            await repl.SendAsync(new SubmitCode("var x = 1;"));
+
+            KernelEvents
```
akshita31: Will this case not produce the "CodeSubmissionEvaluated" event similar to the test "it_notifies_when_submission_is_complete()"

PR #301, opened 2019-06-25
sketching rendering pipeline

colombod:
Render objects to plain text or html


Microsoft.DotNet.Try.Jupyter/Rendering/RendererUtilities.cs
```diff
@@ -0,0 +1,185 @@
...
+                rowsBuffer.AppendLine($"\t\t<td>{keyRendering?.Content ?? string.Empty}</td>");
+                CreateTableRow(memberInfos, engine, element, nullRendering, rowsBuffer);
+                rowsBuffer.AppendLine("\t</tr>");
+            }
+
+            return rowsBuffer.ToString();
+        }
+
+        private static void CreateTableRow(IEnumerable<MemberInfo> memberInfos, IRenderingEngine engine, object element,
+            PlainTextRendering nullRendering, StringBuilder rowsBuffer)
```
akshita31: Can I pass a PlainTextRendering here which is not null ? Both the functions that call it pass a null here, so we can create the nullRendering inside here only ?

WorkspaceServer/Kernel/RenderingEngine.cs
```diff
@@ -0,0 +1,78 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace WorkspaceServer.Kernel
+{
+    public class InClassName
```
akshita31: unused ?

Microsoft.DotNet.Try.Jupyter/Rendering/RendererUtilities.cs
colombod: null rendering is the rendering to use if null. so the caller can decide

Microsoft.DotNet.Try.Jupyter.Tests/PrinterTests.cs
```diff
@@ -0,0 +1,177 @@
...
+            var table = html.DocumentNode.SelectSingleNode("table");
+
+            table.Should().NotBeNull();
+
+            table.SelectNodes("//th")
+                .Select(th => th.InnerText)
+                .Should()
+                .BeEquivalentTo("Url", "Priority");
+
+            table.SelectNodes("//td")
```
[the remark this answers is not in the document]
colombod: is a check for being equivalent, which means same element any order

PR #309, opened 2019-07-03
pound processors

colombod:
code submission can be processed before execution to introduce additional behaviour and the opportunity to manipulate the code submission.


WorkspaceServer/Kernel/ICodeSubmissionPreProcessor.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.CommandLine;
+using System.Threading.Tasks;
+
+namespace WorkspaceServer.Kernel
+{
+    public interface ICodeSubmissionProcessor
```
akshita31: File name doesnt align

Microsoft.DotNet.Try.Jupyter/ExecuteRequestHandler.cs
```diff
@@ -96,13 +96,13 @@ void IObserver<IKernelEvent>.OnNext(IKernelEvent value)
...
 
-        private void OnCodeSubmissionEvaluatedFailed(CodeSubmissionEvaluationFailed codeSubmissionEvaluationFailed, ConcurrentDictionary<Guid, OpenRequest> openRequests)
+        private static void OnCodeSubmissionEvaluatedFailed(CodeSubmissionEvaluationFailed codeSubmissionEvaluationFailed, ConcurrentDictionary<Guid, OpenRequest> openRequests)
         {
             var openRequest = openRequests[codeSubmissionEvaluationFailed.ParentId];
  
             var errorContent = new Error(
                 eName: "Unhandled Exception",
-                eValue: $"{codeSubmissionEvaluationFailed.Error}"
+                eValue: $"{codeSubmissionEvaluationFailed.Message}"
```
akshita31: what will happen to the "Error" property now. Doesnt it have any useful information ?

Microsoft.DotNet.Try.Jupyter/Rendering/DefaultRenderer.cs
```diff
@@ -28,19 +28,24 @@ public IRendering Render(object source, IRenderingEngine engine = null)
...
+            catch (Exception)
+            {
+                return new PlainTextRendering(source?.ToString());
+            }
         }
 
         private string CreateRows(object source, IRenderingEngine engine)
         {
-
-            var props = source.GetType().GetProperties(BindingFlags.Public | BindingFlags.Instance | BindingFlags.Static);
```
akshita31: Just for my understanding, why did we remove the static binding flags?

Microsoft.DotNet.Try.Jupyter/ExecuteRequestHandler.cs
colombod: Error is an object and this this case we use the message (if message is not set then object to string is used or the message field is the error is an exception

Microsoft.DotNet.Try.Jupyter/Rendering/DefaultRenderer.cs
```diff
@@ -28,19 +28,24 @@ public IRendering Render(object source, IRenderingEngine engine = null)
...
+                var table = $@"<table>
     {rows}
 </table>";
 
-            return new HtmlRendering(table);
+                return new HtmlRendering(table);
+            }
+            catch (Exception)
+            {
+                return new PlainTextRendering(source?.ToString());
```
jonsequitur: Examples of exceptions we might see here? Should we be displaying them somewhere?

Issue #307, opened 2019-07-02
[Feature Request] Api for trydotnet session to return the open document


akshita31:
When we create a session using trydotnet, it would be useful to have an api to get the list of the currently opened document, for scenarios like nodatime https://nodatime.org/2.4.x/userguide/recipes where we need to copy code to the active document.

Code for reference 
````
async function setUpTryDotnet() {
            let session = await createTryDotnetSession();

**let openedDocument = session.getOpenDocument(); //propsed api**
            
            let copyTextButton = document.getElementById("trydotnet-copy-button");
            copyTextButton.addEventListener("click", event => setDocumentContent(openedDocument));
        }

        function setDocumentContent(document) {
            //read the content that is present in the window and set the content of the editor
            let contentArea = document.getElementById('trydotnet-content-area');
            document.setContent(contentArea.innerText);
        }

        async function createTryDotnetSession() {
            let configuration = {
                hostOrigin: "http://localhost:58737",
                trydotnetOrigin: "http://localhost:27261",
                useWasmRunner: true
            };

            let project = await trydotnet.createProject({
                packageName :"blazor-console",
                files:[
                        {
                            name: "Program.cs",
                            content:
                                `using System;

namespace fullPaths
{
    class Program
    {
        static void Main(string[] args)
        {
            #region myRegion
            Console.WriteLine("Hello World!");
            #endregion
        }
    }
}`
                        }
                    ]
                    });

            let editors = [document.getElementById("editor")];
            return trydotnet.createSessionWithProjectAndOpenDocument(configuration, editors, window, project, "default.cs");
        }
````

akshita31:
cc @colombod

PR #312, opened 2019-07-03
Get all the open documents from the session

akshita31:
Fixes: https://github.com/dotnet/try/issues/307


Microsoft.DotNet.Try.js/test/session.documentapi.specs.ts
```diff
@@ -77,7 +77,7 @@ describe("A user", () => {
             let document = await session.openDocument({ fileName: "program.cs", region: "controller" });
             document.id().should.equal("program.cs@controller");
             document.getContent().should.equal("//content");
-        });
```
colombod: the lovely ; at the end is gone...

PR #313, opened 2019-07-08
WIP Feature/emit processor

colombod:


WorkspaceServer/Kernel/CodeSubmissionProcessors.cs
```diff
@@ -21,9 +21,10 @@ public class CodeSubmissionProcessors
         public CodeSubmissionProcessors()
         {
             _rootCommand = new RootCommand();
+            _parser = new CommandLineBuilder(_rootCommand).Build();
         }
 
-        public void Register(ICodeSubmissionProcessor processor)
+        public void Add(ICodeSubmissionProcessor processor)
         {
             _processors[processor.Command] = processor;
```
jonsequitur: `Add` seems like a misleading name if this can also replace an existing processor.

WorkspaceServer/Kernel/CodeSubmissionProcessors.cs
```diff
@@ -32,25 +33,37 @@ public void Register(ICodeSubmissionProcessor processor)
...
+                    }
+                    else
+                    {
+                        unhandledLines.Enqueue(currentLine);
+                    }
                 }
+
+                return new SubmitCode(string.Join("\n", unhandledLines), codeSubmission.Id, codeSubmission.ParentId);
+            }
+            catch (Exception e)
```
jonsequitur: This seems overly broad. We should ideally only catch exceptions that we understand, as this might otherwise mask some of our own bugs.

WorkspaceServer/Kernel/RenderingEngine.cs
```diff
@@ -12,21 +12,30 @@ public class RenderingEngine : IRenderingEngine
...
-                return _nullRendering;
+                if (source == null)
+                {
+                    return _nullRendering;
+                }
+
+                var renderer = FindRenderer(source.GetType());
+                return renderer.Render(source, this);
+            }
+            catch (Exception e)
```
jonsequitur: Is there a specific reason (e.g. expected exception type) to do this? `catch Exception` is something that should typically only ever happen at a much higher level in the code.

WorkspaceServer/Kernel/EmitProcessors.cs
```diff
@@ -0,0 +1,64 @@
...
+            Command.Handler = CommandHandler.Create<EmitProcessorsOptions>((options) =>
+            {
+                if (!options.Output.Exists)
+                {
+                    options.Output.Create();
+                }
+                
+                var state = getScriptState();
+
+                var codeFile = new FileInfo(Path.Combine(options.Output.FullName, "code.cs"));
```
jonsequitur: >Path.Combine(options.Output.FullName, "code.cs") [](start = 44, length = 48)

This should be made to work against IDirectoryAccessor from the outset.

WorkspaceServer/Kernel/CodeSubmissionProcessorException.cs
```diff
@@ -0,0 +1,17 @@
...
+
+using System;
+
+namespace WorkspaceServer.Kernel
+{
+    public class CodeSubmissionProcessorException : Exception
+    {
+        public SubmitCode CodeSubmission { get; }
+
+        public CodeSubmissionProcessorException(Exception exception, SubmitCode codeSubmission) : base("CodeSubmission processing failed", exception)
```
jonsequitur: It would also be helpful to refer to the processor that failed.

WorkspaceServer/Kernel/CodeSubmissionProcessors.cs
colombod: changed ot add so we throw exceptions when overriding

WorkspaceServer/Kernel/CodeSubmissionProcessors.cs
colombod: that is wrapping all exceptions so we capture the exception and the code submission

WorkspaceServer/Kernel/EmitProcessors.cs
colombod: indeed

WorkspaceServer/Kernel/RenderingEngine.cs
colombod: this one is wrapping as well, and capturing the evaluation result

WorkspaceServer/Kernel/CodeSubmissionProcessors.cs
jonsequitur: This still obscure the intention. A more honest name would be something like SetProcessorForCommand.

---
In reply to: [301570041](https://github.com/dotnet/try/pull/313#discussion_r301570041) [](ancestors = 301570041)

PR #317, opened 2019-07-10
Integration test 

rchande:


NotIntegrationTests/DotnetTry.cs
```diff
@@ -0,0 +1,91 @@
...
+using System.IO;
+using System.Net.Http;
+using System.Runtime.CompilerServices;
+using System.Text;
+using System.Threading.Tasks;
+using MLS.Agent;
+using MLS.Agent.Tools;
+using WorkspaceServer;
+
+namespace NotIntegrationTests
```
jonsequitur: Well what are they then?

NotIntegrationTests/DotnetTry.cs
```diff
@@ -0,0 +1,91 @@
...
+        {
+            var dotnetTryPackageSource = Environment.GetEnvironmentVariable("dotnetTryPackageSource");
+
+            var directory = !string.IsNullOrWhiteSpace(dotnetTryPackageSource)
+                ? new DirectoryInfo(dotnetTryPackageSource)
+                : new DirectoryInfo(Path.Combine(Path.GetDirectoryName(callerFile), "../artifacts/packages/Debug/Shipping"));
+
+            if (!directory.Exists)
+            {
+                throw new Exception($"Expected packages directory {directory.FullName} to exist but it does not");
```
jonsequitur: Is there a more specific exception that would be appropriate here?

azure-pipelines.yml
```diff
@@ -115,6 +115,13 @@ jobs:
           artifactName: packages
           artifactType: container
 
+      - task: PublishBuildArtifacts@1
```
brettfo: If publishing the tests like this you can remove `<IsPackable>true</>` from `NotIntegrationTests.csproj`.

PR #318, opened 2019-07-10
Should use 4242 if not in try mode

akshita31:
Currently if we launch in try mode using Visual Studio then we launch on 4242, otherwise if we launch from the command line it runs on an arbitrary port and uses https. It would be good if the behavior when we run with F5 is the same as when the tool is run so that if there is an issue when running the tool from the command line, it can also be reproduced using F5 in Visual Studio.

The switching based on mode is also important because as per the settings we have been using so far the agent was assumed to be running on http://localhost:4242 and we should retain that behavior as we change our dockerimage to consume the tool.


akshita31:
Closing this as this is not the intended functionality

PR #326, opened 2019-07-16
Use http in hosted mode

akshita31:
1. Hosted mode scheme should be "http"
2. Try mode scheme should be "https"

### DIFF
--- a/MLS.Agent/Properties/launchSettings.json
+++ b/MLS.Agent/Properties/launchSettings.json
@@ -11,7 +11,6 @@
     "IIS Express": {
       "commandName": "IISExpress",
       "commandLineArgs": "--port 4242 hosted",
-      "launchBrowser": true,
       "environmentVariables": {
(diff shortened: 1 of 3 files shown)


MLS.Agent/Properties/launchSettings.json
```diff
@@ -10,31 +10,29 @@
...
       "commandName": "IISExpress",
-      "commandLineArgs": "hosted",
-      "launchBrowser": true,
+      "commandLineArgs": "--port 4242 hosted",
       "environmentVariables": {
         "ASPNETCORE_ENVIRONMENT": "Development"
       }
     },
-    "MLS.Agent": {
+    "Try": {
```
rchande: Can you add this file to .gitignore? I'm tired of seeing it in PRs.

akshita31: I changed it intentionally in this PR to pass the port option

PR #334, opened 2019-07-19
Command to install the .NET kernel

akshita31:


MLS.Agent/DotnetKernelJupyterInstaller.cs
```diff
@@ -0,0 +1,111 @@
...
+                if(result.ExitCode ==0)
+                {
+                    var dataDirectories = await GetJupyterDataPaths(result.Output);
+                    Installkernel(dataDirectories, console);
+                    console.Out.WriteLine(".NET kernel installation succeded");
+                    return 0;
+                }
+                else
+                {
+                    throw new KernelInstallationFailureException($"Tried to invoke \"jupyter --paths\" but failed with exception: {string.Join("\n", result.Error)}");
```
jonsequitur: It's best to avoid using exceptions for control flow. Since this exception will always be caught and result in the behavior inside your catch block, you can simplify this by just putting the console output and result code inside the else block and getting rid of the `try`/`catch` and exception altogether.

MLS.Agent/DotnetKernelJupyterInstaller.cs
```diff
@@ -0,0 +1,111 @@
...
+                }
+            }
+            catch (KernelInstallationFailureException e)
+            {
+                console.Out.WriteLine($".NET Kernel Installation failed with error {string.Join("\n", e.Message)}");
```
jonsequitur: It's preferable to write to `console.Error`.

MLS.Agent/DotnetKernelJupyterInstaller.cs
```diff
@@ -0,0 +1,111 @@
...
+                            currentPaths.Add(element);
+                        }
+                    }
+                }
+
+                return currentPaths.Select(dir => new DirectoryInfo(dir));
+            }
+            else
+            {
+                throw new KernelInstallationFailureException($"Could not find the jupyter kernel installation directory." +
```
jonsequitur: Rather than throw here, why not just return an empty collection and let the caller determine what action to take?

MLS.Agent/StringExtensions.cs
```diff
@@ -22,5 +23,10 @@ public static IHtmlContent ToHtmlContent(this string value)
         {
             return new HtmlString(value);
         }
+
+        public static string TrimLineEndings(this string value)
+        {
+            return value.Trim('\r', '\n');
```
jonsequitur: `string.Trim` already does this.

MLS.Agent.Tests/DotnetKernelJupyterInstallerTests.cs
```diff
@@ -15,20 +16,25 @@ public class DotnetKernelJupyterInstallerTests
...
+   {dataPath3}
 runtime:
-    C:\Users\akagarw\AppData\Roaming\jupyter\runtime".Split("\n");
+    C:\Users\AppData\Roaming\jupyter\runtime".Split("\n");
 
             var dataDirectories = await DotnetKernelJupyterInstaller.GetJupyterDataPaths(pathsOutput);
             dataDirectories.Should().HaveCount(3);
+            dataDirectories.First().FullName.Should().Be(dataPath1);
+            dataDirectories.Should().Contain(dir => dir.FullName == dataPath1);
+            dataDirectories.Should().Contain(dir => dir.FullName == dataPath1);
```
jonsequitur: These three assertions look redundant.

You can do something like this:

```cs
dataDirectories.Should().BeEquivalentTo(dataPth1, dataPath2, dataPath3);
```

MLS.Agent.Tests/DotnetKernelJupyterInstallerTests.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Threading.Tasks;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace MLS.Agent.Tests
+{
+    public class DotnetKernelJupyterInstallerTests
+    {
+        [Fact]
+        public async Task Should_read_the_jupyter_paths_and_give_the_data_paths()
```
colombod: this is a anaconda related test, what are the paths of a pip based install

MLS.Agent.Tests/DotnetKernelJupyterInstallerTests.cs
```diff
@@ -15,20 +16,25 @@ public class DotnetKernelJupyterInstallerTests
...
 data:
-    C:\Users\akagarw\AppData\Roaming\jupyter
-    C:\Users\akagarw\AppData\Local\Continuum\anaconda3\share\jupyter
-    C:\ProgramData\jupyter
+   {dataPath1}
+   {dataPath2}
+   {dataPath3}
 runtime:
-    C:\Users\akagarw\AppData\Roaming\jupyter\runtime".Split("\n");
+    C:\Users\AppData\Roaming\jupyter\runtime".Split("\n");
```
jonsequitur: This test string should be pulled into a simulator implementation that can be verified against a real Jupyter install.

MLS.Agent/DotnetKernelJupyterInstaller.cs
```diff
@@ -70,29 +70,16 @@ private static void Installkernel(IEnumerable<DirectoryInfo> dataDirectories, IC
...
-                            currentPaths.Add(element);
-                        }
-                    }
-                }
+                var nextHeaderIndex = Array.FindIndex(jupyterPathInfo, dataHeaderIndex + 1, element => element.Trim().EndsWith(":"));
+                if (nextHeaderIndex == -1)
+                    nextHeaderIndex = jupyterPathInfo.Count();
 
-                return currentPaths.Select(dir => new DirectoryInfo(dir));
+                return jupyterPathInfo.Skip(dataHeaderIndex+1).Take(nextHeaderIndex - dataHeaderIndex - 1).Select(dir => new DirectoryInfo(dir.Trim()));
```
jonsequitur: `string.Substring` is much more efficient than using LINQ for string operations.

MLS.Agent/DotnetKernelJupyterInstaller.cs
```diff
@@ -0,0 +1,98 @@
...
+            catch (KernelInstallationFailureException e)
+            {
+                console.Out.WriteLine($".NET Kernel Installation failed with error {string.Join("\n", e.Message)}");
+                return -1;
+            }
+        }
+
+        private static void Installkernel(IEnumerable<DirectoryInfo> dataDirectories, IConsole console)
+        {
+            foreach (var directory in dataDirectories)
```
colombod: why are we creating kernels folder and dumping the kernel spec file and images in all data directories?

MLS.Agent/DotnetKernelJupyterInstaller.cs
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        public static async Task<IEnumerable<DirectoryInfo>> GetJupyterDataPaths(string[] jupyterPathInfo)
+        {
+            var dataHeaderIndex = Array.FindIndex(jupyterPathInfo, element => element.Trim().CompareTo("data:") == 0);
+            if (dataHeaderIndex != -1)
+            {
+                var nextHeaderIndex = Array.FindIndex(jupyterPathInfo, dataHeaderIndex + 1, element => element.Trim().EndsWith(":"));
+                if (nextHeaderIndex == -1)
+                    nextHeaderIndex = jupyterPathInfo.Count();
```
jonsequitur: Please add braces around this.

MLS.Agent/DotnetKernelJupyterInstaller.cs
```diff
@@ -0,0 +1,98 @@
...
+                    {
+                        dotnetkernelDir.Create();
+                    }
+
+                    console.Out.WriteLine($"Installing the .NET kernel in directory: {dotnetkernelDir.FullName}");
+
+                    // Copy the files into the kernels directory
+                    File.Copy("kernels.json", Path.Combine(dotnetkernelDir.FullName, "kernels.json"));
+                    File.Copy("logo-32x32.png", Path.Combine(dotnetkernelDir.FullName, "kernels.json"));
+                    File.Copy("logo-64x64.png", Path.Combine(dotnetkernelDir.FullName, "kernels.json"));
```
jonsequitur: These are all overwriting kernels.json.

MLS.Agent/DotnetKernelJupyterInstaller.cs
```diff
@@ -0,0 +1,98 @@
...
+                    var dotnetkernelDir = kernelDirectory.Subdirectory(".NET");
+                    if (!dotnetkernelDir.Exists)
+                    {
+                        dotnetkernelDir.Create();
+                    }
+
+                    console.Out.WriteLine($"Installing the .NET kernel in directory: {dotnetkernelDir.FullName}");
+
+                    // Copy the files into the kernels directory
+                    File.Copy("kernels.json", Path.Combine(dotnetkernelDir.FullName, "kernels.json"));
```
jonsequitur: Is the working directory expected to contain these files? 

These will need to be extracted from our packaged tool like we do with other resources.

MLS.Agent.Tests/DotnetKernelJupyterInstallerTests.cs
akshita31: Done. Thanks!

MLS.Agent/DotnetKernelJupyterInstaller.cs
rchande: +1, if there's no kernels directory it seems like "something went wrong"

MLS.Agent/DotnetKernelJupyterInstaller.cs
```diff
@@ -0,0 +1,61 @@
...
+            {
+                if (directory.Exists)
+                {
+                    var kernelDirectory = directory.Subdirectory("kernels");
+                    if (!kernelDirectory.Exists)
+                    {
+                        kernelDirectory.Create();
+                    }
+
+                    var dotnetkernelDir = kernelDirectory.Subdirectory(".NET");
```
colombod: path in kernel directory have problems with ```.``` in the name

jonsequitur: This results in a warning but it's not breaking anything. I think this is fine for now.

WorkspaceServer/Paths.cs
```diff
@@ -24,13 +24,16 @@ static Paths()
             NugetCache = String.IsNullOrWhiteSpace(nugetPackagesEnvironmentVariable)
                              ? Path.Combine(UserProfile, ".nuget", "packages")
                              : nugetPackagesEnvironmentVariable;
+
+            JupyterKernelSpecPath = Path.Combine(UserProfile, @"AppData\Local\Continuum\anaconda3\Scripts\jupyter-kernelspec.exe");
```
colombod: this works only in windows with anaconda install.

akshita31:
image

It is working now

MLS.Agent.Tests/FactRunWhenJupyterIsInstalledAttribute.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            this.messageSink = messageSink;
+        }
+
+        public IEnumerable<IXunitTestCase> Discover(
+            ITestFrameworkDiscoveryOptions discoveryOptions,
+            ITestMethod testMethod,
+            IAttributeInfo factAttribute)
+        {
+            if (testMethod.TestClass.Class.Name.Contains("Integration") && !FileSystemJupyterKernelSpec.CheckIfJupyterKernelSpecExists().Result)
```
jonsequitur: It's a good general policy to avoid ever using `Result` as it can often result in deadlocks. I would make `CheckIfJupyterKernelSpecExists` non-async and wrap the async call there in a `Task.Run(...).Wait()`, with a timeout.

akshita31:
image

tested with Linux as well

PR #335, opened 2019-07-19
add up-to-date CI check

brettfo:
Copying from [`dotnet/fsharp`](https://github.com/dotnet/fsharp), adding a CI step to verify we don't unnecessarily rebuild the `node` dependencies.  The process is to build the solution, gather the modification timestamp from some sentinel files, build again, and verify the timestamps haven't changed, indicating a rebuild.

Check the `$FileList` array in `UpToDate.ps1` for the list of sentinel files that we check.

### DIFF
--- a/eng/tests/UpToDate.ps1
+++ b/eng/tests/UpToDate.ps1
@@ -1,0 +1,72 @@
+# This script verifies that subsequent calls to `build.cmd` don't cause js/css to be unnecessarily rebuilt.
+
+[CmdletBinding(PositionalBinding=$false)]
+param (
+    [string][Alias('c')]$configuration = "Debug",
+    [parameter(ValueFromRemainingArguments=$true)][string[]]$properties
+)
+
+Set-StrictMode -version 2.0
+$ErrorActionPreference = "Stop"
+
+try {
+    $RepoRoot = Join-Path $PSScriptRoot ".." | Join-Path -ChildPath ".." -Resolve
+    $BuildScript = Join-Path $RepoRoot "build.cmd"
+
+    # do first build
+    & $BuildScript -configuration $configuration @properties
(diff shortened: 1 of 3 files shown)


brettfo:
Initial run shows a failure (which is good).  Will amend PR with fix before merging.

```
1 of 3 files were re-built.
    D:\a\1\s\MLS.Agent\wwwroot\css\trydotnet.css
##[error]PowerShell exited with code '1'.
```

eng/tests/UpToDate.ps1
```diff
@@ -0,0 +1,72 @@
...
+    & $BuildScript -configuration $configuration @properties
+    if ($LASTEXITCODE -ne 0) {
+        Write-Host "Error running first build."
+        exit 1
+    }
+
+    # list of sentinel files to check
+    $FileList = @(
+        "$RepoRoot\MLS.Agent\wwwroot\api\trydotnet.min.js",
+        "$RepoRoot\MLS.Agent\wwwroot\client\bundle.js",
```
colombod: client can have a number of bundle.js files not just one

brettfo: What's the likelihood of `1.bundle.js` (or any of the others) being re-built and `bundle.js` not?  We can certainly update this to check everything under `wwwroot/`, but I figured I'd initially just check a few files.  Runtime doesn't matter much if we check 3 or 300 files.

brettfo: Turns out there are only 73 files under `wwwroot/` so I just updated it to check them all.

brettfo:
After fixing the `trydotnet.css` issue and triggering a failure on anything changing under `wwwroot/`, the output now ends with:

```
0 of 68 files were re-built.
```

PR #337, opened 2019-07-22
move files and rename to Microsoft.DotNet.Interactive

colombod:


MLS.Agent/CommandLine/CommandLineParser.cs
```diff
@@ -10,6 +10,7 @@
 using System.Threading.Tasks;
 using Clockwise;
 using Microsoft.AspNetCore.Hosting;
+using Microsoft.DotNet.Interactive;
```
jonsequitur: Why does this reference Interactive?

colombod: IKernel is in that namespace

colombod: .AddTransient<IKernel>(c => new CompositeKernel

PR #338, opened 2019-07-22
Condition integration tests on env var RUN_DOTNET_TRY_INTEGRATION_TESTS

rchande:


NotIntegrationTests/ConditionalFact.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
+using Xunit;
+
+namespace NotIntegrationTests
+{
+    public class ConditionalFact : FactAttribute
```
jonsequitur: A more specific name would let people know this is specifically about integration tests.

PR #339, opened 2019-07-22
#r nuget package

rchande:


WorkspaceServer.Tests/Kernel/CompositeKernelTests.cs
```diff
@@ -59,7 +59,7 @@ public async Task When_SubmitCode_command_adds_packages_to_csharp_kernel_then_Pa
                 new CSharpKernel().UseNugetDirective()
             };
 
-            var command = new SubmitCode("#r \"nuget:PocketLogger, 1.2.3\" \nvar a = new List<int>();", "csharp");
+            var command = new SubmitCode("#r \"nuget:Microsoft.Extensions.Logging, 3.0.0-preview6.19304.6\" \nMicrosoft.Extensions.Logging.ILogger logger = null;", "csharp");
```
jonsequitur: You don't need to specify a language if there's only one child kernel.

WorkspaceServer.Tests/PackageRestoreContextTests.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    public class PackageRestoreContextTests
+    {
+        [Fact]
+        public async Task It_works()
+        {
+            using (var context = new PackageRestoreContext())
+            {
+                var refs = await context.AddPackage("FluentAssertions", "5.7.0");
+                refs.ToArray().Length.Should().Be(2);
```
jonsequitur: Why two refs? 

The assert can be more specific.

WorkspaceServer/Kernel/CSharpKernelExtensions.cs
```diff
@@ -5,11 +5,14 @@
 using System.CommandLine.Invocation;
 using Microsoft.DotNet.Interactive;
 using Microsoft.DotNet.Interactive.Events;
+using WorkspaceServer.PackageRestore;
 
 namespace WorkspaceServer.Kernel
 {
     public static class CSharpKernelExtensions
     {
+        private static PackageRestoreContext _restoreContext = new PackageRestoreContext();
```
jonsequitur: Why static?

WorkspaceServer.Tests/PackageRestoreContextTests.cs
```diff
@@ -12,13 +12,19 @@ namespace WorkspaceServer.Tests
...
         {
-            using (var context = new PackageRestoreContext())
-            {
-                var refs = await context.AddPackage("FluentAssertions", "5.7.0");
-                refs.ToArray().Length.Should().Be(2);
-            }
+            var refs = await new PackageRestoreContext().AddPackage("FluentAssertions", "5.7.0");
+            refs.Should()
+                .Contain(r => r.Display.Contains("FluentAssertions.dll")
+                        || r.Display.Contains("System.Configuration.ConfigurationManager"));
```
jonsequitur: Why `||`? This smells of nondeterminism.

WorkspaceServer/Packaging/PackageRestoreContext.cs
```diff
@@ -59,12 +46,16 @@ public async Task<IEnumerable<MetadataReference>> AddPackage(string packageName,
             var dotnet = new Dotnet(package.Directory);
             var result = await dotnet.AddPackage(packageName, packageVersion);
 
+            if (result.ExitCode != 0)
+            {
+                return null;
```
jonsequitur: There's no clear semantic difference between an empty collection and null, so unless we think this is an error condition, `Array.Empty<MetadataReference>` would be better. Otherwise, throw.

WorkspaceServer.Tests/PackageRestoreContextTests.cs
rchande: FluentAssertions nuget package adds these 2 assemblies...

jonsequitur: The assertion will pass if only one is present though. Maybe `Should().BeEquivalentTo()`?

rchande: The overloads there are awkward if you want to use a validation delegate so I've opted to call contains twice.

PR #340, opened 2019-07-22
don't publish build logs during integration run; they're not produced

brettfo:
The Arcade-generated build step `Publish Logs` in the IntegrationTest leg is reporting a warning:

```
Publishing build artifacts failed with an error: Not found PathtoPublish: /home/vsts/work/1/s/artifacts/log/Release
```

While this doesn't fail the build, it does mark it as "Succeeded with issues" and it shows as orange in the build logs; green is much better.

~~Marked [WIP] until I can verify the logs.~~

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -134,54 +134,6 @@ jobs:
     enableTelemetry: true
     helixRepo: dotnet/try
     jobs:
-    - job: IntegrationTest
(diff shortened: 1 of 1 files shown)


brettfo:
Logs look good, no longer WIP.

I had to move the `IntegrationTest` job definition into its own group so I could use a different set of template parameters, but otherwise the job is unchanged.

PR #343, opened 2019-07-23
tracking std out prints

colombod:
doing Console.Write and Console.WriteLine in kernels produces values


WorkspaceServer.Tests/Kernel/CSharpKernelTests.cs
```diff
@@ -192,9 +192,29 @@ public async Task it_produces_values_when_executing_Console_output()
             KernelEvents.OfType<ValueProduced>()
                 .Should()
                 .BeEquivalentTo(
-                    new ValueProduced("value one", kernelCommand, new[] { new FormattedValue(null, "value one"), }),
-                    new ValueProduced("value two", kernelCommand, new[] { new FormattedValue(null, "value two"), }),
-                    new ValueProduced("value three", kernelCommand, new[] { new FormattedValue(null, "value three"), }));
+                    new ValueProduced("value one", kernelCommand, false, new[] { new FormattedValue(null, "value one"), }),
+                    new ValueProduced("value two", kernelCommand, false, new[] { new FormattedValue(null, "value two"), }),
+                    new ValueProduced("value three", kernelCommand, false, new[] { new FormattedValue(null, "value three"), }));
```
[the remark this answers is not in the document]
colombod: myme types are done later on when formatting on raw value is done

Microsoft.DotNet.Interactive/Events/ValueProduced.cs
```diff
@@ -8,16 +8,18 @@ namespace Microsoft.DotNet.Interactive.Events
 {
     public class ValueProduced : KernelEventBase
     {
-        public ValueProduced(
-            object value,
+        public ValueProduced(object value,
             SubmitCode submitCode,
+            bool isLastValue = false,
```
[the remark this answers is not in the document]
colombod: in this case you do know is the latest as it is the last eval of the expression. that tells the framework about doing diplay updated or if this is the output of the cell. A non terminating cell is fine, just not producing an output

WorkspaceServer.Tests/Kernel/CSharpKernelTests.cs
```diff
@@ -178,6 +178,45 @@ public async Task it_aggregates_multiple_submissions()
...
 
+        [Fact]
+        public async Task it_produces_values_when_executing_Console_output()
+        {
+            var kernel = CreateKernel();
+
+            var kernelCommand = new SubmitCode(@"
+Console.Write(""value one"");
+Console.Write(""value two"");
+Console.Write(""value three"");", "csharp");
```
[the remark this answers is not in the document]
colombod: writeLine appends a newline

WorkspaceServer.Tests/Kernel/CSharpKernelTests.cs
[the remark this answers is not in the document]
jonsequitur: I think `FormattedValue` without a mime type is confusing and probably bug prone. I think this test shows that because of the redundancy -- the formatted values ARE the raw values.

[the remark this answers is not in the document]
colombod: null value is generated by the  [Formatter.MimeTypeFor](https://github.com/dotnet/try/blob/0fd2e197e62d54090c591584b3a392bf42fcc7d8/Microsoft.DotNet.Interactive.Rendering/Formatter.cs#L361) method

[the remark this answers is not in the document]
jonsequitur: That's the right behavior. `FormattedValue.MimeType` is intentionally read-only because the expectation was that the mime type must be known at the time a `FormattedValue` is created. It was an oversight that the constructor didn't guard against this.

WorkspaceServer/Servers/Roslyn/TrackingStringWriter.cs
```diff
@@ -0,0 +1,289 @@
...
+{
+    internal class TrackingStringWriter : StringWriter
+    {
+        private class Region
+        {
+            public int Start { get; set; }
+            public int Length { get; set; }
+        }
+         List<Region> _regions = new List<Region>();
+         private bool _trakingSize;
```
rchande: _trackingSize

WorkspaceServer/Servers/Roslyn/TrackingStringWriter.cs
```diff
@@ -0,0 +1,289 @@
...
+
```
rchande: Extra blank line

WorkspaceServer/Servers/Roslyn/TrackingStringWriter.cs
```diff
@@ -0,0 +1,289 @@
...
+
+         public bool WriteOccurred { get; set; }
+
+        public override void Write(char value)
+        {
+            TrackSize(() => base.Write(value));
```
rchande: You're going to reallocate this delegate every time you write, consider caching it.

WorkspaceServer/Servers/Roslyn/TrackingStringWriter.cs
```diff
@@ -0,0 +1,289 @@
...
+
+            _regions.Add(region);
+            action();
+            region.Length = sb.Length - region.Start;
+            _trakingSize = false;
+        }
+        private async Task TrackSizeAsync(Func<Task> action)
+        {
+            WriteOccurred = true;
+            if (_trakingSize)
```
rchande: I feel like this is trying to handle some concurrency scenario but there doesn't seem to be any synchronization?

WorkspaceServer/Servers/Roslyn/TrackingStringWriter.cs
colombod: ops!

WorkspaceServer/Servers/Roslyn/TrackingStringWriter.cs
colombod: is handling re-entrancy, writeLine get decombosed in base class in write(value) and then WriteLine()

Microsoft.DotNet.Interactive.Jupyter/ExecuteRequestHandler.cs
```diff
@@ -152,10 +157,16 @@ private static void OnCodeSubmissionEvaluatedFailed(CodeSubmissionEvaluationFail
...
+                ?new ExecuteResult(
                     openRequest.ExecutionCount,
-                    transient: openRequest.Transient,
+                    transient: transient,
+                    data: valueProduced?.FormattedValues
+                        ?.ToDictionary(k => k.MimeType ?? "text/plain", v => v.Value))
+                : new DisplayData(
+                    transient: transient,
                     data: valueProduced?.FormattedValues
                                        ?.ToDictionary(k => k.MimeType ?? "text/plain", v => v.Value));
```
jonsequitur: We should no longer need a fallback here.

Microsoft.DotNet.Interactive.Rendering/Formatter.cs
```diff
@@ -358,8 +358,7 @@ private static void TryRegisterDefault(string typeName, Action<object, TextWrite
 
         public static string MimeTypeFor(Type type)
         {
-            _mimeTypesByType.TryGetValue(type, out var mimeType);
-            return mimeType;
+           return  _mimeTypesByType.TryGetValue(type, out var mimeType) ? mimeType : "text/plain";
```
jonsequitur: Maybe extract a static property `Default` so we don't keep repeating this string?

Microsoft.DotNet.Interactive/FormattedValue.cs
```diff
@@ -1,13 +1,15 @@
...
+using System;
+
 namespace Microsoft.DotNet.Interactive
 {
     public class FormattedValue
     {
         public FormattedValue(string mimeType, object value)
         {
-            MimeType = mimeType;
+            MimeType = mimeType ?? throw new ArgumentNullException(nameof(mimeType));
```
jonsequitur: This should also check for empty and whitespace, which are also not valid mime types.

PR #355, opened 2019-08-01
fix a couple of bugs, add Tag tests

jonsequitur:


Microsoft.DotNet.Interactive.Rendering.Tests/HtmlFormatterTests.cs
```diff
@@ -241,6 +241,18 @@ public void It_formats_dictionaries_as_tables_with_the_key_on_the_y_axis()
                       .Be(
                           "<table><thead><tr><th><i>key</i></th><th>TypeName</th><th>Id</th></tr></thead><tbody><tr><td>first</td><td>entity one</td><td>123</td></tr><tr><td>second</td><td>entity two</td><td>456</td></tr></tbody></table>");
             }
+
+            [Fact]
+            public void It_formats_arrays_of_disparate_types_good()
```
colombod: will this make PR builds fail?

rendering.sln
```diff
@@ -0,0 +1,48 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
```
colombod: why this solution?

Microsoft.DotNet.Interactive.Rendering.Tests/HtmlFormatterTests.cs
jonsequitur: Ooops, meant to exclude that.

rendering.sln
jonsequitur: Meant to exclude this.

PR #357, opened 2019-08-05
add simple `FSharpKernel` wrapper around fsi.exe

brettfo:
This is more of a proof-of-concept until we decide how we want to properly refactor `fsi`.

Since we're forced to scrape stdin/stdout, all returned values are forced to `text/plain` (hence no formatter support because everything is `System.String`); this will change once we have a proper kernel.

Includes moving most of `CSharpKernelTestBase.cs` into a common `KernelTestBase.cs`.

### DIFF
--- a/MLS.Agent/CommandLine/CommandLineParser.cs
+++ b/MLS.Agent/CommandLine/CommandLineParser.cs
@@ -511,12 +511,11 @@ namespace MLS.Agent.CommandLine
                                 .UseKernelHelpers()
                                 .UseXplot()),
                             PrepareKernel(new FSharpKernel())
-                        };
+                        }.UseExtendDirective();
         }
 
         private static KernelBase PrepareKernel(KernelBase kernel)
         {
-            kernel.UseExtendDirective();
             kernel.SetDefaultRendering();
             return kernel;
         }
--- a/MLS.Agent/CommandLine/CommandLineParser.cs
+++ b/MLS.Agent/CommandLine/CommandLineParser.cs
@@ -14,6 +14,7 @@ using Clockwise;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.DotNet.Interactive;
+using Microsoft.DotNet.Interactive.FSharp;
 using Microsoft.DotNet.Interactive.Jupyter;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
@@ -510,7 +511,8 @@ namespace MLS.Agent.CommandLine
                                 .UseNugetDirective()
                                 .UseExtendDirective()
                                 .UseKernelHelpers()
-                                .UseXplot()
+                                .UseXplot(),
+                            new FSharpKernel()
                         };
         }
     }
--- a/MLS.Agent/CommandLine/CommandLineParser.cs
+++ b/MLS.Agent/CommandLine/CommandLineParser.cs
@@ -506,18 +506,14 @@ namespace MLS.Agent.CommandLine
         {
             return new CompositeKernel
                         {
-                            PrepareKernel(new CSharpKernel()
+                            new CSharpKernel()
+                                .UseDefaultRendering()
                                 .UseNugetDirective()
                                 .UseKernelHelpers()
-                                .UseXplot()),
-                            PrepareKernel(new FSharpKernel())
+                                .UseXplot(),
+                            new FSharpKernel()
+                                .UseDefaultRendering()
                         }.UseExtendDirective();
-        }
-
-        private static KernelBase PrepareKernel(KernelBase kernel)
-        {
-            kernel.SetDefaultRendering();
-            return kernel;
         }
     }
 }
--- a/MLS.Agent/CommandLine/CommandLineParser.cs
+++ b/MLS.Agent/CommandLine/CommandLineParser.cs
@@ -506,14 +506,19 @@ namespace MLS.Agent.CommandLine
         {
             return new CompositeKernel
                         {
-                            new CSharpKernel()
-                                .UseDefaultRendering()
+                            PrepareKernel(new CSharpKernel()
                                 .UseNugetDirective()
-                                .UseExtendDirective()
                                 .UseKernelHelpers()
-                                .UseXplot(),
-                            new FSharpKernel()
+                                .UseXplot()),
+                            PrepareKernel(new FSharpKernel())
                         };
+        }
+
+        private static KernelBase PrepareKernel(KernelBase kernel)
+        {
+            kernel.UseExtendDirective();
+            kernel.SetDefaultRendering();
+            return kernel;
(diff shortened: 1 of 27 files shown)


Microsoft.DotNet.Interactive.FSharp/FSharpKernel.fs
```diff
@@ -0,0 +1,96 @@
...
+    let stdout = StringBuilder()
+    let waitForReady () =
+        async {
+            write <| sprintf ";;printfn \"\\n%s\";;\n" sentinelValue
+            do! Async.AwaitEvent sentinelFound.Publish
+        }
+    let startProcess () =
+        async {
+        if isNull proc then
+            let command = "dotnet"
```
jonsequitur: A couple of classes you could use to reduce some boilerplate here are `WorkspaceServer.Dotnet` (for finding the correct `dotnet` CLI) and `MLS.Agent.Tools.CommandLine` (for managing processes).

WorkspaceServer.Tests/Kernel/CSharpKernelTestBase.cs
```diff
@@ -1,49 +1,21 @@
...
-
-        protected void DisposeAfterTest(IDisposable disposable)
-        {
-            _disposables.Add(disposable);
-        }
-
-        public void Dispose()
-        {
-            _disposables?.Dispose();
+            return new CSharpKernel();
```
jonsequitur: I have some changes in flight here as well and I'm leaning toward always using the same default set of extensions here as we use in the app at runtime. There's really only one supported configuration at the moment so we should test that.

brettfo:
image

Microsoft.DotNet.Interactive.FSharp/FSharpKernel.fs
brettfo: `WorkspaceServer` has a dependency on `Microsoft.DotNet.Interactive.FSharp`, so I can't go the other way and use anything from that assembly. 😢

jonsequitur: Then it's probably appropriate to move the `Dotnet` class into a lower-level project.

brettfo: Right now it's pretty enmeshed in it's current location and it would require moving much more than just that single class (e.g., `MLS.Agent.Tools/(CommandLine|CommandLineResult)` and `WorkspaceServer/PackageSource`) as well as updating all of the references.

I'm all for refactoring this class out, but it feels beyond the scope of this PR.

brettfo: Spoke with @jonsequitur offline and we decided the proper thing to do was to move the `Dotnet` class down to the `MLS.Agent.Tools` assembly.  That's been done as of commit d5560b5053fc17e1e6434de4065cb0fe4be8239a.

WorkspaceServer.Tests/Kernel/CSharpKernelTestBase.cs
brettfo: Currently I think the `.UseDefaultRendering()` extension should stay only in `CSharpKernelTestBase.cs` because the F# kernel doesn't support any of that yet since it's really just a thin wrapper/scraper around `fsi`.

The other alternative is to broaden the extension methods to switch on specific types like so:

``` csharp
public static KernelBase UseDefaultRendering(this KernelBase kernel)
{
    switch (kernel)
    {
        case CSharpKernel csharpKernel:
            csharpKernel.UseDefaultRendering();
            break;
        case FSharpKernel fsharpKernel:
            // noop
            break;
    }

    return kernel;
}

public static KernelBase UseDefaultRendering(this CSharpKernel kernel)
{
    // do C#-specific stuff
}
```

Thoughts?

jonsequitur: Rather than change the type that `UseDefaultRendering` accepts, maybe we should add an overload for `FSharpKernel`. The implication of the proposed method is that it should work for all kernel types, which isn't really possible since we can't know all of the possible implementations of `KernelBase`.

brettfo: I added `void SetDefaultRendering()` to `IKernel` so C# can implement it and F#'s is currently a noop.

Also updated to reflect your recent changes to `master`.

jonsequitur: I don't think that belongs on `IKernel`. Rendering is a fairly separate concern.

MLS.Agent/CommandLine/CommandLineParser.cs
```diff
@@ -467,5 +465,12 @@ Command Verify()
                 return verifyCommand;
             }
         }
+
+        private static KernelBase PrepareKernel(KernelBase kernel)
+        {
+            kernel.UseExtendDirective();
```
jonsequitur: `UseExtendDirective` should be called at the `CompositeKernel` level.

Microsoft.DotNet.Interactive/IKernel.cs
```diff
@@ -16,5 +16,7 @@ public interface IKernel : IDisposable
         IObservable<IKernelEvent> KernelEvents { get; }
 
         Task<IKernelCommandResult> SendAsync(IKernelCommand command, CancellationToken cancellationToken);
+
+        void SetDefaultRendering();
```
jonsequitur: There's no generalization of default rendering that makes sense across all possible kernels, and the concept may not be something we even want to commit to for the two existing kernels. Let's keep the implementations specific to the cases where we understand what this means and keep the interface simpler for now.

MLS.Agent/CommandLine/CommandLineParser.cs
brettfo: Updated.

Microsoft.DotNet.Interactive/IKernel.cs
brettfo: Then move this back to be specific to the C# kernel?  That's where we originally started which seems at odds with how I interpreted [this](https://github.com/dotnet/try/pull/357#discussion_r311773451) comment, specifically, I read this:

> maybe we should add an overload for FSharpKernel.

As

> move this method to KernelBase

But to properly support `CompositeKernel`, the method needed to be on `IKernel`, instead.

jonsequitur: Sorry. I meant that we should have two methods, one for `CSharpKernel` and one for `FSharpKernel`. Their implementations are different and I don't know if this method will make sense for all kernels.

brettfo: Added commit 03e8866, let me know what you think.

brettfo:
Rebased to handle recent changes in `master`.

brettfo:
Ping @jonsequitur for review.

PR #363, opened 2019-08-07
stdio kernel server

rchande:


Microsoft.DotNet.Interactive/Commands/IKernelCommand.cs
```diff
@@ -5,5 +5,6 @@ namespace Microsoft.DotNet.Interactive.Commands
 {
     public interface IKernelCommand
     {
+        string Name { get; }
```
jonsequitur: In the past I've tended to avoid this by letting the serialization layer handle it based on the actual `System.Type`. Within the domain layer, we know the instance's type, so this property isn't useful except in the serialization code.

Microsoft.DotNet.Interactive/Events/IKernelEvent.cs
```diff
@@ -7,6 +7,7 @@ namespace Microsoft.DotNet.Interactive.Events
 {
     public interface IKernelEvent
     {
+        string Type { get; }
```
jonsequitur: Same concern as with `IKernelCommand`.

Microsoft.DotNet.Interactive/KernelStreamClient.cs
```diff
@@ -0,0 +1,87 @@
...
+    {
+
+
+        public Stream Output { get; }
+        public Stream Input { get; }
+
+        private readonly StreamWriter _outputWriter;
+        private readonly IKernel _kernel;
+
+        public KernelStreamClient(IKernel kernel)
```
jonsequitur: Null check please.

WorkspaceServer.Tests/Kernel/CompositeKernelTests.cs
```diff
@@ -163,7 +163,7 @@ public async Task Kernel_can_be_interacted_using_kernel_client()
             var reader = new StreamReader(output, Encoding.UTF8);
 
             var text = reader.ReadToEnd();
-            var events = text.Split("\r\n")
+            var events = text.Split(Environment.NewLine)
```
jonsequitur: We might be running on a different system from where the input was sent, or a notebook authored on one OS might be opened on another, so we should split on both `\n` and `\r\n`. I think we have code elsewhere doing that and it might be a worthwhile utility method to pull out.

jonsequitur: Unrelated, what if the `SubmitCode.Code` contains newlines?

MLS.Agent/CommandLine/CommandLineParser.cs
```diff
@@ -56,6 +56,9 @@ public static class CommandLineParser
             StartServer startServer = null,
             InvocationContext context = null);
 
+        public delegate Task<int> KernelServer(
```
jonsequitur: Maybe `StartKernelServer` or something? We're more typically using verb names for delegate types.

WorkspaceServer.Tests/Kernel/CompositeKernelTests.cs
rchande: A single command/response must not contain newlines because we read them by calling console.ReadLine() (effectively). The code we're looking at here is the test code, which just pushes everything into a single stream and then reads it all back at once (hence the splitting)

MLS.Agent/CommandLine/KernelServerCommand.cs
```diff
@@ -0,0 +1,27 @@
...
+namespace MLS.Agent.CommandLine
+{
+    public static class KernelServerCommand
+    {
+        public static async Task<int> Do(
+            IConsole console)
+        {
+            var kernel = new CompositeKernel()
+            {
+                  new CSharpKernel()
```
jonsequitur: We should share the code that sets up the kernel extensions, etc. with the Jupyter mode startup.

WorkspaceServer.Tests/Kernel/CompositeKernelTests.cs
jonsequitur: What's the process when someone submits a cell with multiple lines of code inside it? If we read the input to the end, instead of line by line, could it simplify this case?

Microsoft.DotNet.Interactive/Events/UnrecognizedCommand.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.DotNet.Interactive.Commands;
+
+namespace Microsoft.DotNet.Interactive.Events
+{
+    internal class UnrecognizedCommand : IKernelEvent
```
jonsequitur: `CommandNotRecognized` would be more in line with event naming conventions. Although I also think this is a developer error condition, so an exception or error channel would be more appropriate.

Microsoft.DotNet.Interactive/StreamKernelEvent.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Microsoft.DotNet.Interactive
+{
+    public class StreamKernelEvent
+    {
+        public int Id { get; set; }
+        public string Type { get; set; }
```
jonsequitur: For consistency this should be `EventType` or else `StreamKernelCommand.CommandType` should be renamed to `Type`.

Microsoft.DotNet.Interactive/StreamKernelExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    public static class StreamKernelExtensions
+    {
+        static int id = 0;
+
+        public static int WriteMessage(this StreamWriter writer, IKernelCommand command)
+        {
+            var message = new StreamKernelCommand()
+            {
+                Id = id++,
```
jonsequitur: Any concurrency concerns here?

WorkspaceServer.Tests/Kernel/CompositeKernelTests.cs
rchande: I think it's not valid json to have an unescaped line break in a string. It's fair for us to expect clients to escape their newlines.

rchande: Will add a test for this though

PR #364, opened 2019-08-08
update install script to be arcade-aware

brettfo:
This script assumes (incorrectly?  I'm open to discussion) that `.\restore.cmd` has already been run.


brettfo:
I'm committing (ha!) a sin and merging this with failing tests because this script isn't used during CI.

PR #365, opened 2019-08-08
Add xplot support

akshita31:
This PR adds a KernelExtension which is loaded by default for now. On loading this extension adds a formatter for the "PlotlyChart". This formatter produces a javascript which looks something like this:

image


akshita31:
cc @eerhardt @pgovind

WorkspaceServer/WorkspaceServer.csproj
```diff
@@ -107,6 +107,7 @@
     <ProjectReference Include="..\Microsoft.DotNet.Try.Project\Microsoft.DotNet.Try.Project.csproj" />
     <ProjectReference Include="..\MLS.Agent.Tools\MLS.Agent.Tools.csproj" />
     <ProjectReference Include="..\Microsoft.DotNet.Try.Protocol\Microsoft.DotNet.Try.Protocol.csproj" />
+    <ProjectReference Include="..\XPlot.DotNet.Interactive.KernelExtensions\XPlot.DotNet.Interactive.KernelExtensions.csproj" />
```
jonsequitur: Thoughts on putting this in MLS.Agent rather than here?

WorkspaceServer.Tests/Kernel/XplotKernelExtensionTests.cs
```diff
@@ -0,0 +1,101 @@
...
+                 ));
+        }
+
+        [Fact]
+        public async Task When_a_chart_is_returned_the_value_produced_hash_the_plotly_newPlot_call()
+        {
+            var kernel = CreateKernel();
+            kernel.UseDefaultExtensions();
+
+            await kernel.SendAsync(new SubmitCode("#r nuget:XPlot.Plotly"));
```
jonsequitur: I think for now this shouldn't be needed (although we'll need to modify the ScriptOptions in `CSharpKernel`). Like `PocketView` and the display helper, it should be available by default.

WorkspaceServer/WorkspaceServer.csproj
akshita31: What is the advantage of one over another ? Right now all the kernel specific stuff lives in WorkspaceServer right ?

XPlot.DotNet.Interactive.KernelExtensions/XPlotKernelExtension.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            Formatter<PlotlyChart>.Register((chart, writer) =>
+           {
+               writer.Write(GetChartHtml(chart));
+           }, "text/html");
+
+            return Task.CompletedTask;
+        }
+
+        private string GetChartHtml(PlotlyChart chart)
```
jonsequitur: This should return `IHtmlContent`.

WorkspaceServer.Tests/Kernel/XplotKernelExtensionTests.cs
akshita31: In that case things will work in c# out of the box and not for fsharp unless we explicitly add that dependency for f# as well. Do we need a list of references "somewhere" and those should be added to both f# and c# by default ?

jonsequitur: @brettfo Thoughts?

MLS.Agent/KernelExtensions.cs
```diff
@@ -0,0 +1,21 @@
...
+
+using Clockwise;
+using Microsoft.DotNet.Interactive;
+using XPlot.DotNet.Interactive.KernelExtensions;
+
+namespace MLS.Agent
+{
+    public static class KernelExtensions
+    {
+        public static T UseDefaultExtensions<T>(this T kernel)
```
jonsequitur: The method name doesn't convey what it's actually doing. Can we make it more specific?

XPlot.DotNet.Interactive.KernelExtensions/XPlotKernelExtension.cs
```diff
@@ -0,0 +1,47 @@
...
+            html.Append(@"
+require.config({paths:{plotly:'https://cdn.plot.ly/plotly-latest.min'}});
+require(['plotly'], function(Plotly) { 
+");
+
+            html.Append(chartHtml, scriptStart + 1, scriptEnd - scriptStart - 1);
+
+            html.AppendLine(@"});");
+
+            html.Append(chartHtml.AsSpan().Slice(scriptEnd).ToString());
```
eerhardt: You should use the same StringBuilder overloads as you have above - (string, start, length), and not create a span and call ToString on it, since that creates and copies unnecessary memory.

WorkspaceServer/Kernel/CSharpKernel.cs
```diff
@@ -66,7 +66,8 @@ private void SetupScriptOptions()
                     typeof(Task<>).Assembly,
                     typeof(IKernel).Assembly, 
                     typeof(CSharpKernel).Assembly, 
-                    typeof(PocketView).Assembly);
+                    typeof(PocketView).Assembly,
+                    typeof(XPlot.Plotly.PlotlyChart).Assembly);
```
eerhardt: Do we really need this reference by default? What if we left it up to the user to `#r "nuget:XPlot"` if they want?

jonsequitur: We can go either way. Initially though, since the extension is referenced by the tool at build time, and we haven't improved the assembly loader, I'm concerned that `#r nuget` will run us afoul of the multiple assembly load bug.

XPlot.DotNet.Interactive.KernelExtensions/XPlotKernelExtension.cs
jonsequitur: Even simpler is to pass in the `TextWriter` from the `Formatter.Register` call on line 15.

WorkspaceServer/Kernel/CSharpKernel.cs
eerhardt: Ok, that seems reasonable to me, for now.

XPlot.DotNet.Interactive.KernelExtensions/XPlotKernelExtension.cs
eerhardt: Even better. In that case, I'd suggest making this project target `netcoreapp2.1` or higher (or `netstandard2.1`), and then take advantage of the new `ReadOnlySpan` overload - https://docs.microsoft.com/en-us/dotnet/api/system.io.textwriter.write?view=netcore-2.1#System_IO_TextWriter_Write_System_ReadOnlySpan_System_Char__.

I'm sure it won't make a dramatic different performance wise, but why make the computer do unnecessary work when it doesn't have to.

MLS.Agent.Tests/XplotKernelExtensionTests.cs
```diff
@@ -0,0 +1,98 @@
...
+                .OfType<ValueProduced>()
+                .Should().
+                ContainSingle(valueProduced =>
+                    valueProduced.FormattedValues.Any(formattedValue =>
+                       formattedValue.Value.ToString().Contains("require.config({paths:{plotly:\'https://cdn.plot.ly/plotly-latest.min\'}});")
+                 ));
+        }
+
+        [Fact]
+        public async Task When_a_chart_is_returned_the_value_produced_hash_the_plotly_newPlot_call()
```
rchande: hash -> has

MLS.Agent.Tests/XplotKernelExtensionTests.cs
```diff
@@ -0,0 +1,98 @@
...
+
+            await kernel.SendAsync(new SubmitCode("using XPlot.Plotly;"));
+            await kernel.SendAsync(new SubmitCode("new PlotlyChart()"));
+
+            KernelEvents
+                .ValuesOnly()
+                .OfType<ValueProduced>()
+                .Should().
+                ContainSingle(valueProduced =>
+                    valueProduced.FormattedValues.Any(formattedValue =>
```
rchande: Based on the test name, should this be asserting html mimetype as well?

MLS.Agent.Tests/XplotKernelExtensionTests.cs
```diff
@@ -0,0 +1,98 @@
...
+
+            await kernel.SendAsync(new SubmitCode("using XPlot.Plotly;"));
+            await kernel.SendAsync(new SubmitCode("new PlotlyChart()"));
+
+            KernelEvents.ValuesOnly()
+                .OfType<ValueProduced>()
+                .Should().
+                ContainSingle(valueProduced =>
+                    valueProduced.FormattedValues.Any(formattedValue =>
+                        formattedValue.MimeType == "text/html"
```
rchande: I think it would be more precise to test the contents and mimetype against the same formattedValue instance.

MLS.Agent/KernelExtensions.cs
rchande: This reads pretty clearly to me as "UseXPlot"

XPlot.DotNet.Interactive.KernelExtensions/XPlotKernelExtension.cs
```diff
@@ -0,0 +1,47 @@
...
+
+            StringBuilder html = new StringBuilder(chartHtml.Length);
+            html.Append(chartHtml,0, scriptStart);
+
+            html.Append(@"
+require.config({paths:{plotly:'https://cdn.plot.ly/plotly-latest.min'}});
+require(['plotly'], function(Plotly) { 
+");
+
+            html.Append(chartHtml, scriptStart + 1, scriptEnd - scriptStart - 1);
```
rchande: We already have code that uses HtmlAgilityPack. Consider using that to parse the document in a less brittle way?

XPlot.DotNet.Interactive.KernelExtensions/XPlotKernelExtension.cs
rchande: Because this function is just chart -> string, I suspect you can write a really precise unit test covering it

rchande: And then your kernel level tests just need to test the actual integration

akshita31: Thanks I did this in the commit here - https://github.com/dotnet/try/pull/365/commits/29fad696321476bed1b5ce5efc442921f2e63ae7

MLS.Agent.Tests/XplotKernelExtensionTests.cs
```diff
@@ -35,63 +42,57 @@ public async Task When_a_chart_is_returned_the_value_produced_has_html_with_the_
...
                     valueProduced.FormattedValues.Any(formattedValue =>
-                       formattedValue.Value.ToString().Contains("require.config({paths:{plotly:\'https://cdn.plot.ly/plotly-latest.min\'}});")
-                 ));
+                    formattedValue.MimeType == "text/html" &&
+                       formattedValue.Value.ToString().Contains("require([\'plotly\'], function(Plotly)")
+                       && formattedValue.Value.ToString().Contains("require.config({paths:{plotly:\'https://cdn.plot.ly/plotly-latest.min\'}});")
+                 )) ;
         }
 
-        [Fact]
```
rchande: Move to new file please

MLS.Agent.Tests/XplotKernelExtensionTests.cs
```diff
@@ -35,63 +42,57 @@ public async Task When_a_chart_is_returned_the_value_produced_has_html_with_the_
...
-        public async Task When_a_chart_is_returned_the_value_produced_hash_the_plotly_newPlot_call()
+        public class GetChartHtmlTests
```
rchande: Move to new file please

XPlot.DotNet.Interactive.KernelExtensions/XPlotKernelExtension.cs
akshita31: Done now - https://github.com/dotnet/try/pull/365/commits/64fa37565a268dd37c7c447699df37abfb1bc6cd Thanks!

XPlot.DotNet.Interactive.KernelExtensions/XPlotKernelExtension.cs
akshita31: Instead of parsing strings, I have modified the code to use the HTMLAgilityPack

MLS.Agent.Tests/XplotKernelExtensionTests.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Threading.Tasks;
+using WorkspaceServer.Tests.Kernel;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace MLS.Agent.Tests
+{
+    public partial class XplotKernelExtensionTests : CSharpKernelTestBase
+    {
+
```
rchande: Extra blank line

XPlot.DotNet.Interactive.KernelExtensions/XPlotKernelExtension.cs
```diff
@@ -0,0 +1,55 @@
...
+using static Microsoft.DotNet.Interactive.Rendering.PocketViewTags;
+
+namespace XPlot.DotNet.Interactive.KernelExtensions
+{
+    public class XPlotKernelExtension : IKernelExtension
+    {
+        public Task OnLoadAsync(IKernel kernel)
+        {
+            Formatter<PlotlyChart>.Register((chart, writer) =>
+           {
```
rchande: Weird indenting

PR #380, opened 2019-08-16
Fix verify command

akshita31:
1. If no directory has been speicified verify should take the current directory in the VerifyOptions
2. The directory from the VerifyOptions must be used inside the VerifyCommand to create a package registry and get a directory accessor, so that it is ensured that all operations are performed with respect to the dir passed in the verify options

### DIFF
--- a/MLS.Agent.Tests/CommandLine/VerifyCommandTests.cs
+++ b/MLS.Agent.Tests/CommandLine/VerifyCommandTests.cs
@@ -92,8 +92,7 @@ This is some sample code:
                 await VerifyCommand.Do(
                     new VerifyOptions(root),
                     console,
-                    () => directoryAccessor,
-                    PackageRegistry.CreateForTryMode(root));
+                    (dir) => directoryAccessor);
 
                 console.Out
                     .ToString()
@@ -121,8 +120,7 @@ This is some sample code:
                 await VerifyCommand.Do(
                     new VerifyOptions(root),
                     console,
-                    () => directoryAccessor,
-                    PackageRegistry.CreateForTryMode(root));
+                    (dir) => directoryAccessor);
 
                 _output.WriteLine(console.Out.ToString());
 
@@ -156,8 +154,7 @@ This is some sample code:
                 await VerifyCommand.Do(
                     new VerifyOptions(root),
                     console,
-                    () => directoryAccessor,
-                    PackageRegistry.CreateForTryMode(root));
+                    (dir) => directoryAccessor);
 
                 _output.WriteLine(console.Out.ToString());
 
@@ -198,8 +195,7 @@ public class EmptyClassTwo {}
                 var resultCode = await VerifyCommand.Do(
                     new VerifyOptions(root),
                     console,
-                    () => directoryAccessor,
-                    PackageRegistry.CreateForTryMode(root));
+                    (dir) => directoryAccessor);
 
                 _output.WriteLine(console.Out.ToString());
 
@@ -230,8 +226,7 @@ public class EmptyClassTwo {}
                 var resultCode = await VerifyCommand.Do(
                     new VerifyOptions(root),
                     console,
-                    () => directoryAccessor,
-                    PackageRegistry.CreateForTryMode(root));
+                    (dir) => directoryAccessor);
 
                 _output.WriteLine(console.Out.ToString());
 
@@ -258,8 +253,7 @@ public class EmptyClassTwo {}
                 var resultCode = await VerifyCommand.Do(
                     new VerifyOptions(root),
                     console,
-                    () => directoryAccessor,
-                    PackageRegistry.CreateForTryMode(root));
+                    (dir) => directoryAccessor);
 
                 _output.WriteLine(console.Out.ToString());
 
@@ -286,8 +280,7 @@ public class EmptyClassTwo {}
                 var resultCode = await VerifyCommand.Do(
                                      new VerifyOptions(rootDirectory),
                                      console,
-                                     () => directoryAccessor,
-                                     PackageRegistry.CreateForTryMode(rootDirectory));
+                                     (dir) => directoryAccessor);
 
                 _output.WriteLine(console.Out.ToString());
 
@@ -328,8 +321,7 @@ public class Program
                 var resultCode = await VerifyCommand.Do(
                                      new VerifyOptions(rootDirectory),
                                      console,
-                                     () => directoryAccessor,
-                                     PackageRegistry.CreateForTryMode(rootDirectory));
+                                     (dir) => directoryAccessor);
 
                 _output.WriteLine(console.Out.ToString());
 
@@ -349,8 +341,7 @@ public class Program
                 var resultCode = await VerifyCommand.Do(
                                      new VerifyOptions(rootDirectory),
                                      console,
-                                     () => directoryAccessor,
-                                     PackageRegistry.CreateForTryMode(rootDirectory));
+                                     (dir) => directoryAccessor);
 
                 console.Error.ToString().Should().Contain("No markdown files found");
                 resultCode.Should().NotBe(0);
@@ -383,8 +374,7 @@ public class Program
                 var resultCode = await VerifyCommand.Do(
                                      new VerifyOptions(rootDirectory),
                                      console,
-                                     () => directoryAccessor,
-                                     PackageRegistry.CreateForTryMode(rootDirectory));
+                                     (dir) => directoryAccessor);
 
                 console.Out.ToString().Should().Contain("Session cannot span projects or packages: --session one");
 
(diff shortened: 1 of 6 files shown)


MLS.Agent.Tests/CommandLine/VerifyCommandTests.cs
```diff
@@ -718,8 +699,7 @@ public async Task When_standalone_markdown_with_non_editable_code_block_targetin
                 var resultCode = await VerifyCommand.Do(
                     new VerifyOptions(root),
                     console,
-                    () => directoryAccessor,
-                    PackageRegistry.CreateForTryMode(root),
+                    (dir) => directoryAccessor,
```
rchande: Consider calling this `_` or `unused`

WorkspaceServer/Packaging/Package.cs
```diff
@@ -575,7 +575,25 @@ protected virtual bool ShouldDoFullBuild()
...
+            if (DesignTimeBuildResult != null)
+            {
+                var filesInDirectory = Directory.GetFiles("*.cs").Select(file => file.FullName);
+                var documentsInLastBuild = DesignTimeBuildResult.GetWorkspace().CurrentSolution.Projects.First().Documents.Select(document => document.FilePath);
+                if (filesInDirectory.Count() != documentsInLastBuild.Count())
+                {
+                    return true;
+                }
+
+                return !documentsInLastBuild.SequenceEqual(filesInDirectory);
```
rchande: I don't think you can assume that the documents are represented in the same order in the workspace and in the file system.

WorkspaceServer.Tests/PackageTests.cs
```diff
@@ -185,5 +186,17 @@ public async Task CanSupportBlazor_indicates_whether_the_package_supports_Wasm_r
             var package = await registry.Get<ICanSupportWasm>(packageName);
             package.CanSupportWasm.Should().Be(expected);
         }
+
+        [Fact]
+        public async Task If_new_files_are_added_after_one_successful_build_the_new_workspace_includes_the_file()
```
rchande: Should probably cover the other case you added in the up to date detection (file count is same but an existing file is gone)

MLS.Agent/CommandLine/VerifyCommand.cs
```diff
@@ -23,11 +23,11 @@ public static class VerifyCommand
...
             VerifyOptions options,
             IConsole console,
-            Func<IDirectoryAccessor> getDirectoryAccessor,
-            PackageRegistry packageRegistry,
+            Func<DirectoryInfo, IDirectoryAccessor> getDirectoryAccessor,
             StartupOptions startupOptions = null)
         {
-            var directoryAccessor = getDirectoryAccessor();
+            var packageRegistry = PackageRegistry.CreateForTryMode(options.Dir);
+            var directoryAccessor = getDirectoryAccessor(options.Dir);
```
rchande: This pattern is also a little weird--could we push the directoryAccessor into the options?

akshita31: The options are used at the command line and they take the directory from the base `dotnet try` command. I was unable to determine how to convert the dirArgument from the base command to a DirectoryAccessor for the subcommand. There is a method called TryConvertArgument, but I couldnt figure out how to use it to get the desired result. Maybe @jonsequitur might have more ideas.

For now, I have used the dir from the VerifyOptions and then created a DirectoryAccessor which is passed on to the VerifyCommand, hence VerifyCommand now takes only a directory accessor and no directory.

MLS.Agent.Tests/CommandLine/VerifyCommandTests.cs
akshita31: Not using that lambda now

WorkspaceServer.Tests/PackageTests.cs
akshita31: I just realised that this up-to-date check will cause a serious performance regression for the Web scenario(or the non-rebuildable packages case).

akshita31: Moved the up-to-date check in the constructor of the rebuildable package which is the one used by verify and also `dotnet try`

MLS.Agent.Tests/CommandLine/CommandLineParserTests.cs
```diff
@@ -59,9 +60,9 @@ public CommandLineParserTests(ITestOutputHelper output)
                     _installPackageSource = options.AddSource;
                     return Task.CompletedTask;
                 },
-                verify: (options, console, startupOptions) =>
+                verify: (directoryAccessor, console, startupOptions) =>
```
jonsequitur: How does this `directoryAccessor` relate to `startupOptions.Dir`? This seems like we have a concept being duplicated. Ideally we'd use `IDirectoryAccessor` (and the same instance) for both.

MLS.Agent.Tests/CommandLine/CommandLineParserTests.cs
```diff
@@ -27,7 +28,7 @@ public class CommandLineParserTests : IDisposable
         private PackOptions _packOptions;
         private InstallOptions _installOptions;
         private PackageSource _installPackageSource;
-        private VerifyOptions _verifyOptions;
```
jonsequitur: Is the `VerifyOptions` class still being used? Thoughts on changing its `Dir` property to be a `IDirectoryAccessor` and keeping the options pattern?

MLS.Agent.Tests/CommandLine/CommandLineParserTests.cs
```diff
@@ -340,7 +341,14 @@ public async Task Verify_argument_specifies_root_directory()
         {
             var directory = Path.GetDirectoryName(typeof(VerifyCommand).Assembly.Location);
             await _parser.InvokeAsync($"verify {directory}", _console);
-            _verifyOptions.Dir.FullName.Should().Be(directory);
+            RelativePath.NormalizeDirectory(_verifyDirectoryAccessor.GetFullyQualifiedRoot().FullName).Should().Be(RelativePath.NormalizeDirectory(directory));
```
jonsequitur: Is it necessary to call `NormalizeDirectory` on both sides of the assertion? It makes the test a little harder to read and I'm curious about why they wouldn't match.

MLS.Agent/CommandLine/VerifyCommand.cs
jonsequitur: Ideally `PackageRegistry.CreateForTryMode` would also take in `IDirectoryAccessor` and not `DirectoryInfo`.

MLS.Agent.Tests/CommandLine/CommandLineParserTests.cs
akshita31: I have added an extension method to do the same

WorkspaceServer/FileSystemDirectoryAccessor.cs
```diff
@@ -16,6 +17,10 @@ public class FileSystemDirectoryAccessor : IDirectoryAccessor
     {
         private readonly DirectoryInfo _rootDirectory;
 
+        public FileSystemDirectoryAccessor(string value): this(new DirectoryInfo(value))
```
jonsequitur: It would be good to rename `value` to something more informative.

WorkspaceServer/PackageRegistry.cs
```diff
@@ -157,7 +157,7 @@ private Task<IPackage> GetPackageFromPackageBuilder(string packageName, Budget b
             });
         }
 
-        public static PackageRegistry CreateForTryMode(DirectoryInfo project, PackageSource addSource = null)
+        public static PackageRegistry CreateForTryMode(IDirectoryAccessor directoryAccessor, PackageSource addSource = null)
```
jonsequitur: Maybe `projectDirectory` would be a more informative parameter name? 

Naming the parameter after the type isn't adding information. We know it's an `IDirectoryAccessor`. The parameter name can help explain what it will be used for in this context.

akshita31: Done. Thanks

WorkspaceServer/FileSystemDirectoryAccessor.cs
akshita31: Done in https://github.com/dotnet/try/pull/380/commits/85257c62d3b42fd47d7ba43cec53ce16eb056f37

MLS.Agent.Tests/CommandLine/CommandLineParserTests.cs
akshita31: Changed the verify options and the startup options to only use the directory accessor

MLS.Agent.Tests/CommandLine/CommandLineParserTests.cs
akshita31: Modified the command line parser to input a  directory and create a directory accessor which is used in the StartupOptions and the VerifyOptions

PR #385, opened 2019-08-20
xplot working in jupyter and nteract

colombod:
Handle plotly loading via require js  in both jupyter and nteract

### DIFF
--- a/XPlot.DotNet.Interactive.KernelExtensions/XPlotKernelExtension.cs
+++ b/XPlot.DotNet.Interactive.KernelExtensions/XPlotKernelExtension.cs
@@ -6,6 +6,7 @@ using System.Text;
 using System.Threading.Tasks;
 using XPlot.Plotly;
 using static Microsoft.DotNet.Interactive.Rendering.PocketViewTags;
+using Html = Microsoft.DotNet.Interactive.Rendering.Html;
 
 namespace XPlot.DotNet.Interactive.KernelExtensions
 {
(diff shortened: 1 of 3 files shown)


XPlot.DotNet.Interactive.KernelExtensions/XPlotKernelExtension.cs
```diff
@@ -30,24 +31,37 @@ public string GetChartHtml(PlotlyChart chart)
...
+    script.setAttribute(""src"", ""https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js""); 
+    script.onload = function(){
+        renderPlotly();
+    };
+    document.getElementsByTagName(""head"")[0].appendChild(script); 
+}
+else {
+    renderPlotly();
+}");
             newScript.AppendLine("</script>");
```
jonsequitur: Why not use `PocketView` rather than add a dependency on HtmlAgilityPack?

colombod: we need parsing anyway to select the div and script element from the xplot html

jonsequitur: Can't they just be rendered as separate script elements?

colombod: The script tag injects the other script element conditionally if requrirejs is not defined, they cannot be 2 tags

PR #388, opened 2019-08-21
enable display update

colombod:
display api returns a displayHandle that can be used to update the output in teh frontend


Microsoft.DotNet.Interactive/Commands/DisplayValue.cs
```diff
@@ -5,11 +5,13 @@ namespace Microsoft.DotNet.Interactive.Commands
...
     {
-        public DisplayValue(FormattedValue formattedValue)
+        public DisplayValue(FormattedValue formattedValue, string id = null)
         {
             FormattedValue = formattedValue;
+            Id = id;
         }
 
         public FormattedValue FormattedValue { get; }
+        public string Id { get; }
```
jonsequitur: For consistency, `DisplayId` would be better.

Microsoft.DotNet.Interactive/Commands/UpdateDisplayValue.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.DotNet.Interactive.Commands
+{
+    public class UpdateDisplayValue : KernelCommandBase
```
jonsequitur: Maybe `UpdateDisplayedValue`?

Microsoft.DotNet.Interactive/Commands/DisplayValue.cs
jonsequitur: Although I think this identifies a "value", not a "display", so here an in `UpdateDisplayValue` maybe we should rename this to `ValueId`?

Microsoft.DotNet.Interactive/Events/ValueProduced.cs
```diff
@@ -11,17 +11,25 @@ public class ValueProduced : KernelEventBase
         public ValueProduced(object value,
             IKernelCommand command,
             bool isLastValue = false,
-            IReadOnlyCollection<FormattedValue> formattedValues = null) : base(command)
+            IReadOnlyCollection<FormattedValue> formattedValues = null,
+            string id= null,
```
jonsequitur: `ValueId`?

Microsoft.DotNet.Interactive/Events/ValueProduced.cs
```diff
@@ -11,17 +11,25 @@ public class ValueProduced : KernelEventBase
...
 
         public object Value { get; }
 
         public bool IsLastValue { get; }
 
         public IReadOnlyCollection<FormattedValue> FormattedValues { get; }
+
+        public string Id { get; }
+
+        public bool IsUpdatedValue { get; }
```
jonsequitur: It seems like this should implicitly be true if an id is set. Would it not be a coding error if this is set to `true` while `id` is set to `null`?

Microsoft.DotNet.Interactive/IDisplayHandle.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.DotNet.Interactive
+{
+    public interface IDisplayHandle
+    {
+        void Update(object value);
```
jonsequitur: Maybe `UpdateDisplayedValue`?

Microsoft.DotNet.Interactive/Commands/UpdateDisplayedValue.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.DotNet.Interactive.Commands
+{
+    public class UpdateDisplayedValue : KernelCommandBase
+    {
+        public UpdateDisplayedValue(FormattedValue formattedValue, string valueId = null)
```
eerhardt: `valueId` is required, right? I don't think it should default to `null`.

Microsoft.DotNet.Interactive/KernelBase.cs
```diff
@@ -190,6 +198,45 @@ command switch
...
+                        valueId: displayedValue.ValueId,
+                        isUpdatedValue:true)
+                    );
+
+                context.OnCompleted();
+
+                return Task.CompletedTask;
+            };
+
+            displayedValue.Handler = invocationContext =>
```
eerhardt: Doesn't this just over-write the Handler set in the line above this?

WorkspaceServer.Tests/Kernel/CSharpKernelRenderingTests.cs
```diff
@@ -99,6 +99,30 @@ public async Task Display_helper_can_be_called_without_specifying_class_name()
...
+
+            formatted
+                .Should()
+                .ContainSingle(v =>
+                    v.MimeType == "text/html" &&
+                    v.Value.ToString().Contains("<b>hello</b>"))
+                .And
+                .ContainSingle(v =>
+                    v.MimeType == "text/html" &&
+                    v.Value.ToString().Contains("<b>world</b>"));
```
jonsequitur: This assertion doesn't check that they're also in the correct order.

Microsoft.DotNet.Interactive.Jupyter/ExecuteRequestHandler.cs
```diff
@@ -78,10 +78,9 @@ public async Task Handle(JupyterRequestContext context)
             }
         }
 
-        private static Dictionary<string, object> CreateTransient()
+        private static Dictionary<string, object> CreateTransient(string displayId)
```
jonsequitur: `valueId`

Microsoft.DotNet.Interactive/Events/ValueProduced.cs
colombod: is false if id is not set, but can be false even if Id is set

Microsoft.DotNet.Interactive/Events/ValueProduced.cs
```diff
@@ -15,6 +16,11 @@ public class ValueProduced : KernelEventBase
             string valueId= null,
             bool isUpdatedValue = false) : base(command)
         {
+            if (isUpdatedValue && valueId == null)
+            {
+                throw new ArgumentException("isUpdatedValue cannot be true with a null valueId", nameof(valueId));
```
jonsequitur: Please use `nameof` for the referenced parameter names.

Issue #300, opened 2019-06-24
Insert.md from `dotnet try demo` seems outdated


Falco20019:
#### Describe the bug
I did a `dotnet try verify .` on a fresh `dotnet try demo` folder. It complains a lot about the `Insert.md` that seems not to be part of the quick start documentation but is in [docs/GettingStarted](https://github.com/dotnet/try/blob/master/docs/GettingStarted/Insert.md) and therefore also delivered in the generated `demo.zip`.

#### Did this error occur while using `dotnet try` or online?
- [x] `dotnet-try`
- [ ] online

#### What kind of error was it?
- [ ] User Interface (UI):  For example the output never displayed 
- [ ] Service Error: For example "The service is temporarily unavailable. We are working on it"
- [x] Other: Outdated files

#### Screenshots
![grafik](https://user-images.githubusercontent.com/940619/60020254-f23bb600-968f-11e9-8236-0652eea56ddd.png)

#### Please complete the following:
 - OS
    - [x] Windows 10
    - [ ] macOS
    - [ ] Linux (Please specify distro)
    - [ ] iOS
    - [ ] Android

PR #391, opened 2019-08-21
Verify should pass for the folder created by the demo command

akshita31:
Fixes: https://github.com/dotnet/try/issues/300

1. Renamed Insert.md--> ReadOnlySnippets.md
2. ReadOnlySnippets should use a --project argument so that verify passes against those and users are able to execute it when using `dotnet try demo`


docs/GettingStarted/QuickStart.md
```diff
@@ -5,6 +5,7 @@
 - [Show snippets using regions](./Regions.md)
 - [Verify your Project](./Verify.md)
 - [Passing Arguments](./PassingArgs.md)
+- [Using Read-only Snippets](./ReadOnlySnippets.md)
```
jonsequitur: You'll want to make an equivalent change in the other files that also have these nav links.

PR #392, opened 2019-08-21
[WIP] Magic

jonsequitur:
This enables the directive APIs to implement IPython-style magic commands. 

I've changed the experimental `#kernel csharp` directive to be a cell-scoped language selector, enabling, e.g.:

```csharp
%%csharp

var x = 123;
``` 

In this example "csharp" is actually the subkernel name, so this particular magic will be enabled automatically when a kernel is added to the `CompositeKernel`.


jonsequitur:
image

Issue #395, opened 2019-08-22
Adding extensions to the dotnet kernel using nuget packages


akshita31:
**What is an extension ?**
An extension is anything that can be used to modify the behavior of the kernel without necessarily being checked into the kernel code. For example, if we would like the kernel to print "Hello World" after every successful code submission, it can be performed via adding an extension to the kernel.

**Possible applications of an extension**
1. Adding custom formatters for various types. For example, we can register a formatter for a dictionary which will display the keys and values in the dictionary in a table format with the keys in red color.
2. Support for various visualisation libraries. The Xplot Support that is present right now is done via the kernel extension. The kernel extension when loaded registers a formatter for the XplotChart type.
3. Support for magic commands like `%%javascript`. Such an extension would possible add a middleware that will read the code submitted and check if the magic command is present and do the evaluation accordingly.

**Existing implementation**
Right now an extension can be added to the kernel using the `#extend path_to_dll` directive. The problem with this approach is the acquisition story. Once extension author create extensions, how do they make their extension easily available to the notebook users.

**Proposed Implementation**
The extension can be a part of a nuget package. When a user does 
``` #r nuget:MyNugetPackage```
We look for a specific path inside the Nuget Package, say `/Extensions` and try to find all the implementations of `IKernelExtensions` and load them into the kernel.

This approach can also be used to load scripts into the notebook context. Similar to the "Extensions", there can also be a "/Scripts" folder inside the nuget package containing the script files. We look for all those script files and pass `#load MyScript.csx` into the underlying Script engine.

akshita31:
cc @eerhardt

jonsequitur:
We might want to differentiate the types of extensions by using specific paths.

Following the example of Roslyn analyzers (which are found under `{package_root}\analyzers\dotnet\cs`), we might want to differentiate by language. For example:

`{package_root}\interactive-extensions\dotnet\cs` -> Assemblies containing extensions for `CSharpKernel`
`{package_root}\interactive-extensions\dotnet\fs` -> Assemblies containing extensions for `FSharpKernel`

But language-agnostic extensions would also be possible: 

`{package_root}\interactive-extensions\dotnet` -> Assemblies containing extensions for `CompositeKernel`

On another dimension, we've also discussed supporting extension via scripts, which, if found in the conventional package location, would be loaded using `#load`. This might look something like:

`{package_root}\interactive-extensions\scripts\dotnet\cs` -> `#load {script_path}` sent to `CSharpKernel`
[...]

PR #396, opened 2019-08-23
Load extensions using #r nuget  directive

akshita31:
Addresses #395.


Microsoft.DotNet.Interactive/KernelBase.cs
```diff
@@ -89,25 +88,16 @@ command switch
...
-                                     .Where(t => typeof(IKernelExtension).IsAssignableFrom(t))
-                                     .ToArray();
-
-                foreach (var extensionType in extensionTypes)
-                {
-                    var extension = (IKernelExtension) Activator.CreateInstance(extensionType);
-
-                    await extension.OnLoadAsync(pipelineContext.Kernel);
-                }
+                await ExtensionLoader.LoadFromAssembly(assembly, pipelineContext.Kernel);
```
jonsequitur: There's a bit of confusion in these methods right now but this should use `context` rather than `pipelineContext` (though they're likely the same instance).

WorkspaceServer.Tests/NuGetPackageExtensionLoaderTest.cs
```diff
@@ -0,0 +1,94 @@
...
+namespace WorkspaceServer.Tests
+{
+    public class NuGetPackageExtensionLoaderTest : CSharpKernelTestBase
+    {
+        public NuGetPackageExtensionLoaderTest(ITestOutputHelper output) : base(output)
+        {
+        }
+
+        [Fact]
+        public async Task Given_a_directory_loads_the_extensions_in_extensions_folder()
```
jonsequitur: I think the name `It_loads_the_extensions_in_extensions_folder` is simpler. The "given" wording tends to signal a BDD-style test, which this is not.

WorkspaceServer/Packaging/NuGetPackageExtensionLoader.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Linq;
+using System.Reflection;
+using System.Runtime.Loader;
+using System.Threading.Tasks;
+
+namespace WorkspaceServer.Packaging
+{
+    public class NuGetPackageExtensionLoader
+    {
+        public NuGetPackageExtensionLoader()
```
jonsequitur: This constructor is unnecessary since it's implicit.

Microsoft.DotNet.Interactive/ExtensionLoader.cs
```diff
@@ -0,0 +1,28 @@
...
+using System;
+using System.Linq;
+using System.Reflection;
+using System.Threading.Tasks;
+
+namespace Microsoft.DotNet.Interactive
+{
+    public static class ExtensionLoader
+    {
+        public static async Task LoadFromAssembly(Assembly assembly, IKernel kernel)
```
jonsequitur: This can be `internal`.

WorkspaceServer/FileSystemDirectoryAccessor.cs
```diff
@@ -99,6 +100,13 @@ public IEnumerable<RelativeFilePath> GetAllFilesRecursively()
                                     new RelativeFilePath(PathUtilities.GetRelativePath(_rootDirectory.FullName, f.FullName)));
         }
 
+        public IEnumerable<RelativeFilePath> GetFiles(string searchPattern)
```
jonsequitur: We shouldn't add this method.

WorkspaceServer/Packaging/NuGetPackageExtensionLoader.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+        }
+
+        public async Task LoadExtensionsInDirectoryAsync(FileSystemDirectoryAccessor rootDirectory, IKernel kernel)
```
jonsequitur: This should take `IDirectoryAccessor`.

Microsoft.DotNet.Interactive/KernelBase.cs
akshita31: Done

WorkspaceServer/FileSystemDirectoryAccessor.cs
akshita31: Deleted it now. Thanks

WorkspaceServer/Packaging/NuGetPackageExtensionLoader.cs
akshita31: Split the thing into two, one which takes a `IDirectoryAccessor` and looks for dlls inside the extensions folder, and one which takes a assemblypath and loads the assembly present at that path

Microsoft.DotNet.Interactive/ExtensionLoader.cs
akshita31: If we make this internal, how will it be accessible from the tests

WorkspaceServer.Tests/NugetPackageKernelExtensionFinderTests.cs
```diff
@@ -0,0 +1,34 @@
...
+
+namespace WorkspaceServer.Tests
+{
+    public class NugetPackageKernelExtensionFinderTests
+    {
+        [Fact]
+        public void Gets_all_extensions_dlls_inside_nuget_package_folder()
+        {
+            var nugetpackageName = "myNuget";
+            var directory = new DirectoryInfo(Directory.GetCurrentDirectory()).Subdirectory(nugetpackageName);
```
jonsequitur: Any reason to use `Directory.GetCurrentDirectory` rather than hard-coding a dummy path?

Microsoft.DotNet.Interactive/Commands/LoadExtensionInNuGetPackage.cs
```diff
@@ -0,0 +1,23 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.IO;
+
+namespace Microsoft.DotNet.Interactive.Commands
+{
+    public class LoadExtensionInNuGetPackage : KernelCommandBase
```
jonsequitur: `LoadExtensionFromNuGetPackage` would be a little clearer.

Microsoft.DotNet.Interactive/Commands/LoadExtensionInNuGetPackage.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        public LoadExtensionInNuGetPackage(NugetPackageReference nugetPackageReference, IEnumerable<FileInfo> metadataReferences)
```
jonsequitur: Please add null checks for these.

Microsoft.DotNet.Interactive/Events/NuGetPackageAdded.cs
```diff
@@ -1,11 +1,13 @@
...
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using Microsoft.DotNet.Interactive.Commands;
+
 namespace Microsoft.DotNet.Interactive.Events
 {
     public class NuGetPackageAdded : KernelEventBase
     {
-        public NuGetPackageAdded(NugetPackageReference packageReference)
+        public NuGetPackageAdded(AddNugetPackage addNugetPackage,  NugetPackageReference packageReference): base(addNugetPackage)
```
jonsequitur: Please add a null check.

Microsoft.DotNet.Interactive/ExtensionLoaded.cs
```diff
@@ -0,0 +1,18 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.DotNet.Interactive.Commands;
+using Microsoft.DotNet.Interactive.Events;
+
+namespace Microsoft.DotNet.Interactive
+{
+    public class ExtensionLoaded : KernelEventBase
+    { 
+        public ExtensionLoaded(string extensionPath)
```
jonsequitur: If we think about display feedback, I wonder if something a little richer than a string would be useful. What other use cases might there be for this event? What can subscribers do with a string?

Microsoft.DotNet.Interactive/KernelBase.cs
```diff
@@ -84,29 +85,47 @@ command switch
                             context,
                             next),
 
+                        LoadExtensionInNuGetPackage loadCSharpExtension =>
```
jonsequitur: Is this is C#-specific? If so, it should go into the `CSharpKernel`. This is worth discussing with @TIHan.

Microsoft.DotNet.Interactive/KernelExtensionPathResolver.cs
```diff
@@ -0,0 +1,29 @@
...
+using System.IO;
+using MLS.Agent.Tools;
+
+namespace Microsoft.DotNet.Interactive
+{
+    public class KernelExtensionPathResolver
+    {
+        public static bool TryGetExtensionPath(DirectoryInfo directory, IKernel handlingKernel, out DirectoryInfo extensionDirectory)
+        {
+            if(handlingKernel.Name == "csharp")
```
jonsequitur: It's possible to name a `CSharpKernel` differently, so this would be more robust if you check the type of the kernel.

Microsoft.DotNet.Interactive/KernelExtensionPathResolver.cs
```diff
@@ -0,0 +1,29 @@
...
+            {
+                extensionDirectory = directory.Subdirectory("interactive-extensions/cs");
+                return true;
+            }
+
+            else if(handlingKernel.Name == "fsharp")
```
jonsequitur: Same.

Microsoft.DotNet.Interactive/KernelBase.cs
TIHan: This is something we need to have a discussion with @KevinRansom about as he has been working on #r nuget for F#. It was mentioned that we could and probably should share so that there isn't a difference in behavior between two implementations of resolving nuget packages for the interactive experience.

I imagine that F# by itself could have its own way of resolving nuget packages, but through an API allow a custom handler.

Microsoft.DotNet.Interactive/KernelExtensionPathResolver.cs
akshita31: I am running into the layering problem again. The Csharp kernel lives in the WorkspaceServer and we cant reference it from the `Microsoft.Dotnet.Interactive` assembly.

jonsequitur: One possible approach then might be to ask the `IKernel` what its path convention is instead of hard-coding it. This could be done by adding an interface, something like `IExtensibleKernel`, because not all `IKernel`s might support extensions via NuGet. 

(I think we'll also want to move `CSharpKernel` into a new project, `Microsoft.DotNet.Interactive.CSharp`, but that doesn't have to happen in this PR).

Microsoft.DotNet.Interactive/KernelExtensionPathResolver.cs
```diff
@@ -0,0 +1,29 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.IO;
+using MLS.Agent.Tools;
+
+namespace Microsoft.DotNet.Interactive
+{
+    public class KernelExtensionPathResolver
+    {
+        public static bool TryGetExtensionPath(DirectoryInfo directory, IKernel handlingKernel, out DirectoryInfo extensionDirectory)
```
jonsequitur: This is fairly vague contract. Should it be a relative or absolute path? How does the caller know what path to pass in?

WorkspaceServer.Tests/Kernel/KernelExtensionLoaderTests.cs
```diff
@@ -0,0 +1,92 @@
...
+
+        }
+
+        private static async Task<FileInfo> CreateExtensionDll(DirectoryInfo baseDirectory, DirectoryInfo extensionOutputDirectory)
+        {
+            var extensionCodeDirectory = baseDirectory.CreateSubdirectory("extensionCode");
+            var microsoftDotNetInteractiveDllPath = typeof(IKernelExtension).Assembly.Location;
+            var extensionName = baseDirectory.Name;
+
+            new InMemoryDirectoryAccessor(extensionCodeDirectory)
```
jonsequitur: Please refactor so we don't have multiple copies of this code.

akshita31: Done. Thanks!

Microsoft.DotNet.Interactive/ExtensionLoaded.cs
```diff
@@ -0,0 +1,19 @@
...
+
+using Microsoft.DotNet.Interactive.Commands;
+using Microsoft.DotNet.Interactive.Events;
+using System.IO;
+
+namespace Microsoft.DotNet.Interactive
+{
+    public class ExtensionLoaded : KernelEventBase
+    { 
+        public ExtensionLoaded(FileInfo extensionPath)
```
jonsequitur: This should have a null check.

Microsoft.DotNet.Interactive/KernelExtensionPathResolver.cs
akshita31: Modified to return a directoryaccessor

Microsoft.DotNet.Interactive/IExtensibleKernel.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using MLS.Agent.Tools;
+
+namespace Microsoft.DotNet.Interactive
+{
+    public interface IExtensibleKernel
+    {
+        RelativeDirectoryPath ExtensionsPath { get; }
```
jonsequitur: An interface should expose behavior, not state. 

A specific thing to consider about exposing a path as state is that a path is just a means to get hold of an assembly or script,  so it's more relevant to the loader, not the kernel. And keep in mind there's more than one path that might be valid for a given kernel (e.g. an assembly path and a script path both for C# extensions.)

akshita31: Now the IExtensibleKernel defines a method that can load extensions inside a given directory

Microsoft.DotNet.Interactive/IExtensibleKernel.cs
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using MLS.Agent.Tools;
+using System.Threading.Tasks;
+
+namespace Microsoft.DotNet.Interactive
+{
+    public interface IExtensibleKernel
+    {
+        public Task LoadExtensionsInDirectory(IDirectoryAccessor directory, KernelInvocationContext invocationContext);
```
jonsequitur: Will there always be an invocation context? What about extensions that we want to load on startup?

Microsoft.DotNet.Interactive/KernelBase.cs
```diff
@@ -60,55 +62,73 @@ private void AddDirectiveMiddlewareAndCommonCommandHandlers()
...
+            {
+                if (invocationContext.HandlingKernel is IExtensibleKernel extensibleKernel)
+                {
+                    if (NuGetPackagePathResolver.TryGetNuGetPackageBasePath(loadExtensionFromNuGetPackage.NugetPackageReference, loadExtensionFromNuGetPackage.MetadataReferences, out var nugetPackageDirectory))
+                    {
+                        await extensibleKernel.LoadExtensionsInDirectory(nugetPackageDirectory, context);
+                    }
+                }
+
+                context.Complete();
```
jonsequitur: You shouldn't call complete if you plan to call the continuation as this will prevent later steps in the pipeline from publishing events. `C

WorkspaceServer.Tests/Kernel/CSharpKernelTests.cs
```diff
@@ -407,59 +409,43 @@ public async Task The_extend_directive_can_be_used_to_load_a_kernel_extension()
...
+        }
+
+
+        [Fact]
+        public async Task Should_load_extension_in_nuget_package()
+        {
+            var directory = Create.EmptyWorkspace().Directory;
+
+            const string nugetPackageName = "myNugetPackage";
+            var nugetPackageDirectory = new FileSystemDirectoryAccessor(directory.Subdirectory($"{nugetPackageName}/2.0.0"));
```
jonsequitur: Can we test this using `InMemoryDirectoryAccessor`?

WorkspaceServer/Kernel/CSharpKernel.cs
```diff
@@ -23,17 +23,19 @@
...
-    public class CSharpKernel : KernelBase
+    public class CSharpKernel : KernelBase, IExtensibleKernel
     {
         internal const string KernelName = "csharp";
 
         private static readonly MethodInfo _hasReturnValueMethod = typeof(Script)
             .GetMethod("HasReturnValue", BindingFlags.Instance | BindingFlags.NonPublic);
 
         protected CSharpParseOptions ParseOptions = new CSharpParseOptions(LanguageVersion.Default, kind: SourceCodeKind.Script);
+        public RelativeDirectoryPath ExtensionsPath { get; }
```
jonsequitur: Any reason this needs to be public?

WorkspaceServer/Kernel/CSharpKernel.cs
```diff
@@ -46,6 +48,7 @@ public CSharpKernel()
             _metadataReferences = ImmutableArray<MetadataReference>.Empty;
             SetupScriptOptions();
             Name = KernelName;
+            ExtensionsPath = new RelativeDirectoryPath("interactive-extensions/cs");
```
jonsequitur: Will both scripts and assemblies be under this path?

Microsoft.DotNet.Interactive/KernelBase.cs
jonsequitur: I don't think we need a different way to resolve a NuGet package but might need custom handling once it's resolved, if only for different extension paths within the package.

Microsoft.DotNet.Interactive/KernelBase.cs
```diff
@@ -60,55 +62,73 @@ private void AddDirectiveMiddlewareAndCommonCommandHandlers()
...
         }
 
+        private async Task HandleLoadExtensionInNuGetPackage(
+            LoadExtensionFromNuGetPackage loadExtensionFromNuGetPackage,
+            KernelInvocationContext invocationContext,
+            KernelPipelineContinuation next)
+        {
+            loadExtensionFromNuGetPackage.Handler = async context =>
+            {
+                if (invocationContext.HandlingKernel is IExtensibleKernel extensibleKernel)
```
jonsequitur: If the handling kernel is not an extensible kernel, it should be an error, no?

WorkspaceServer.Tests/Kernel/CSharpKernelTests.cs
akshita31: This is an end to end test which tests that given a nuget package dll, is it able to load the extension from that, and since "LoadAssembly" is being called, it needs the artifacts to actually exist on disk.

WorkspaceServer/Kernel/CSharpKernel.cs
akshita31: I renamed it to say "AssemblyExtensionsPath"

WorkspaceServer/Kernel/CSharpKernel.cs
akshita31: No, made it private now. Thanks

Microsoft.DotNet.Interactive/KernelBase.cs
```diff
@@ -60,55 +62,73 @@ private void AddDirectiveMiddlewareAndCommonCommandHandlers()
...
+                            HandleLoadExtensionInNuGetPackage(
+                                loadCSharpExtension,
+                                context,
+                                next),
 
                             _ => next(command, context)
                         });
         }
 
+        private async Task HandleLoadExtensionInNuGetPackage(
```
jonsequitur: You can rename the method to match the command type.

Microsoft.DotNet.Interactive/KernelExtensionLoader.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.IO;
+using System.Linq;
+using System.Runtime.Loader;
+using System.Threading.Tasks;
+
+namespace Microsoft.DotNet.Interactive
+{
+    public class KernelExtensionLoader
+    {
+        public async Task<bool> TryLoadFromAssembly(FileInfo assemblyFile, IKernel kernel)
```
jonsequitur: `assemblyFile` should have a null check.

Microsoft.DotNet.Interactive/KernelBase.cs
akshita31: Yes, I removed the call to next since it is the only place that command needs to be handled.

Microsoft.DotNet.Interactive/KernelBase.cs
akshita31: Yes added a CommandFailed event if that occurs

Microsoft.DotNet.Interactive/KernelBase.cs
akshita31: Done

Microsoft.DotNet.Interactive/IExtensibleKernel.cs
akshita31: We need to "Publish" an event saying that the extension has been loaded and we need the context for that. If we would like to load extensions at startup, how would we "Publish" that event if we dont have the KernelInvocationContext ?

I think once we have that scenario, it would not be a hard change to make this method maybe return a list of extensions that it loaded(so that it can be called at startup) and then the caller can do whatever necessary. Thoughts ?

Microsoft.DotNet.Interactive/KernelExtensionLoader.cs
```diff
@@ -0,0 +1,62 @@
...
+                                   .Where(t => typeof(IKernelExtension).IsAssignableFrom(t))
+                                   .ToArray();
+
+            foreach (var extensionType in extensionTypes)
+            {
+                var extension = (IKernelExtension)Activator.CreateInstance(extensionType);
+                await extension.OnLoadAsync(kernel);
+            }
+
+            if (extensionTypes.Count() > 0)
```
jonsequitur: It's better to use `Length` than `Count()` here. The latter is a LINQ method and triggers iteration, while the former is a property of the array.

You can also simplify the branching logic to:

```csharp
return extensionTypes.Length > 0;
```

akshita31: Done. Thanks

Microsoft.DotNet.Interactive/KernelExtensionLoader.cs
```diff
@@ -27,7 +27,7 @@ public async Task<bool> TryLoadFromAssembly(FileInfo assemblyFile, IKernel kerne
             var assembly = AssemblyLoadContext.Default.LoadFromAssemblyPath(assemblyFile.FullName);
             var extensionTypes = assembly
                                    .ExportedTypes
-                                   .Where(t => typeof(IKernelExtension).IsAssignableFrom(t))
+                                   .Where(t => typeof(IKernelExtension).IsAssignableFrom(t) && !t.IsInterface)
```
jonsequitur: There are other types that can't be instantiated: abstract classes and generic type definitions.

This same logic can be found here and could be refactored to an extension on `Type`, something like `CanBeInstantiated`:

https://github.com/dotnet/try/blob/f79ef5cc022b81e024eac68927eff83f9f25355d/Microsoft.DotNet.Interactive/CommandDeserializer.cs#L22-L25

akshita31: Done. Thanks

MLS.Agent.Tools/TypeExtensions.cs
```diff
@@ -16,5 +16,13 @@ public static string ReadManifestResource(this Type type, string resourceName)
                 return reader.ReadToEnd();
             }
         }
+
+        public static bool CanBeInstantiatedFrom(this Type type, Type resultantType)
```
jonsequitur: This can be decomposed a little. Assignability is a different concern from instantiability so should probably not be included in the method.

PR #400, opened 2019-08-27
Move the DirectoryAccessor and related classes to MLS.Agent.Tools

akshita31:


Microsoft.DotNet.Interactive/Microsoft.DotNet.Interactive.csproj
```diff
@@ -7,6 +7,7 @@
     <!-- Clockwise isn't strongly signed -->
   </PropertyGroup>
   <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis" Version="3.1.0-beta3-final" />
```
colombod: why do we need a dependency on this package?

colombod:
> @colombod There are tests for the `Relativepath` and the `FileSystemDirectoryAccessor` but they live in the WorkspaceServerTests.csproj and I think it is fine to leave them there for now and not create a separate test project for those.Thoughts ?

If code is in  project A and is only tested because consumed by project B I think this opens to a lot of troubles. if the code as moved then probably things like the simulator tests should be in a test project that targets project A.

PR #413, opened 2019-08-29
Load native dependencies

rchande:


WorkspaceServer.Tests/Kernel/CSharpKernelTests.cs
```diff
@@ -464,5 +465,75 @@ public async Task OnLoadAsync(IKernel kernel)
...
+
+        [Fact]
+        public async Task Loads_native_dependencies_from_nugets()
+        {
+            var kernel = new CompositeKernel
+            {
+                new CSharpKernel().UseNugetDirective(new NativeAssemblyLoadHelper())
+            };
+
+            var command = new SubmitCode(@"#r ""nuget:Microsoft.ML, 1.3.1""
```
jonsequitur: @eerhardt Can you suggest a minimal and fast code submission to verify that the native assembly was correctly loaded?

WorkspaceServer.Tests/Kernel/CompositeKernelTests.cs
```diff
@@ -254,8 +254,7 @@ public async Task Kernel_can_pound_r_nuget_using_kernel_client()
 
             var input = new MemoryStream();
             var writer = new StreamWriter(input, Encoding.UTF8);
-            writer.WriteMessage(new SubmitCode("#kernel csharp"));
-            writer.WriteMessage(new SubmitCode(@"#r nuget:""Microsoft.Extensions.Logging"""));
+            writer.WriteMessage(new SubmitCode(@"#r nuget:""Microsoft.Spark, 0.4.0"""));
```
colombod: the string should be #r ""nuget:Microsoft.Spark, 0.4.0"""

WorkspaceServer.Tests/Kernel/CSharpKernelTests.cs
```diff
@@ -464,5 +465,75 @@ public async Task OnLoadAsync(IKernel kernel)
...
+            events
+                .Should()
+                .ContainSingle(e => e is NuGetPackageAdded);
+
+            events
+                .Should()
+                .ContainSingle(e => e is CodeSubmissionEvaluated);
+
+            events.Should()
+                .Contain(e => e is DisplayedValueProduced && ((DisplayedValueProduced)e).Value as string == "success\r\n");
```
jonsequitur: What's the significance of the `\r\n` in the output?

WorkspaceServer.Tests/Kernel/CSharpKernelTests.cs
```diff
@@ -464,5 +465,75 @@ public async Task OnLoadAsync(IKernel kernel)
...
+
+        MLContext mlContext = new MLContext();
+        var pipeline = mlContext.Transforms
+            .Concatenate(""Features"", ""SepalLength"", ""SepalWidth"", ""PetalLength"", ""PetalWidth"")
+            .Append(mlContext.Clustering.Trainers.KMeans(""Features"", numberOfClusters: 2));
+
+try
+{
+    pipeline.Fit(mlContext.Data.LoadFromEnumerable(data));
+    Console.WriteLine(""success"");
```
colombod: display will generate the event you are looking for and is more realistic then using Console.WriteLine

WorkspaceServer.Tests/Kernel/CSharpKernelTests.cs
```diff
@@ -464,5 +465,75 @@ public async Task OnLoadAsync(IKernel kernel)
...
+}
+catch (Exception e)
+{
+    Console.WriteLine(e);
+}", "csharp");
+
+            var result = await kernel.SendAsync(command);
+
+            var events = result.KernelEvents
```
jonsequitur: This could hang forever.

I added functionality to make this kind of test easier. Here's an example:

https://github.com/dotnet/try/blob/7d1dd593bab0172a981b7f8618960dc105ffdae1/Microsoft.DotNet.Interactive.Jupyter.Tests/MagicCommandTests.cs#L42

rchande: I don't think this works for commands that do async work? Kernel.SendAsync returning doesn't mean the command has done anything yet. 

Note that I did try this and my test fails because the assertions run before anything has happened.

MLS.Agent/NativeAssemblyLoadHelper.cs
```diff
@@ -0,0 +1,57 @@
...
+                {
+                    NativeLibrary.SetDllImportResolver(args.LoadedAssembly, Resolve);
+                }
+            };
+        }
+
+        private IntPtr Resolve(string libraryName, Assembly assembly, DllImportSearchPath? searchPath)
+        {
+            var basePath = Path.GetDirectoryName(assembly.Location);
+            var nativeAssembly = Path.Combine(basePath, "..", "..", "runtimes", _tfm, "native", libraryName) + ".dll";
```
eerhardt: This should really be using the NuGet native assets in the `project.assets.json`, and not assuming this path.

For one thing, this isn't taking into account the RID-fallback. You can have native assets in a `debian-x64` folder, and those will be used on any debian based system. They don't have to be just in `linux-x64`.

WorkspaceServer.Tests/Kernel/CSharpKernelTests.cs
eerhardt: If you want something light-weight, maybe not using `Microsoft.ML` and using a different package that has native assemblies?  One that I know of https://www.nuget.org/packages/LibGit2Sharp. That actually is a PERFECT test case, if you look at it's `runtimes` folders, it hits exactly the issue I was referring to above:

image

eerhardt: If you want to continue using ML.NET, the below code is probably the fastest thing I can think of from a public API perspective. Clustering 4 rows of 4 floats is super fast.

MLS.Agent/NativeAssemblyLoadHelper.cs
eerhardt: Or even better would be the `.deps.json` file, but I assume you aren't getting one since you aren't going through the full SDK/MSBuild process.

WorkspaceServer.Tests/Kernel/CSharpKernelTests.cs
jonsequitur: If that's the case it's safest to provide a timeout.

jonsequitur: Actually, this should work in any case because the events are available on the `KernelEvents` observable even after they were dispatched, precisely to make this sort of thing easier.

MLS.Agent/NativeAssemblyLoadHelper.cs
rchande: Yes, the plan is to parse deps.json, this is basically some demoware based on partner requests :)

MLS.Agent/NativeAssemblyLoadHelper.cs
```diff
@@ -0,0 +1,49 @@
...
+        private AssemblyDependencyResolver _resolver;
+
+        public NativeAssemblyLoadHelper()
+        {
+        }
+
+        public void Configure(string path)
+        {
+            if (_resolver != null)
+                return;
```
jonsequitur: Nit: curly braces.

PR #414, opened 2019-08-30
Kernel Interrupt api

colombod:


Microsoft.DotNet.Interactive/Commands/InterruptKernel.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace Microsoft.DotNet.Interactive.Commands
 {
-    public class InterruptExecution : KernelCommandBase
+    public class InterruptKernel : KernelCommandBase
```
jonsequitur: The thing being interrupted isn't the kernel. It's an operation or an invocation or something.

Microsoft.DotNet.Interactive.Jupyter.Tests/InterruptRequestHandlerTests.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.DotNet.Interactive.Jupyter.Tests
+{
+    public class InterruptRequestHandlerTests
+    {
+
+    }
```
jonsequitur: So tested!

Microsoft.DotNet.Interactive.Jupyter.Tests/InterruptRequestHandlerTests.cs
```diff
@@ -0,0 +1,66 @@
...
+        {
+            var kernel = new CSharpKernel();
+            var handler = new InterruptRequestHandler(kernel);
+            var request = Message.Create(new DisplayData(), null);
+            Func<Task> messageHandling = () => handler.Handle(new JupyterRequestContext(_serverChannel, _ioPubChannel, request, _kernelStatus));
+            messageHandling.Should().ThrowExactly<InvalidOperationException>();
+        }
+
+        [Fact]
+        public async Task handles_InterruptRequest()
```
akshita31: Should we merge this one and the one below into one test as this one doesnt have any assertion ?

colombod: it would throw if not using the right message

akshita31: Ok. Thanks!

PR #415, opened 2019-08-30
command serializer to  output camlCase as events

colombod:
serialized kernel command are serialized with camlCase convention


Microsoft.DotNet.Interactive/StreamKernelCommand.cs
```diff
@@ -1,9 +1,19 @@
...
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Newtonsoft.Json;
+
+namespace Microsoft.DotNet.Interactive
 {
     public class StreamKernelCommand
     {
+        [JsonProperty("id")]
```
jonsequitur: Do we need both this and the `CamelCasePropertyNamesContractResolver`?

colombod: the contract resolver is to force the properties of nested objects

jonsequitur: Understood, but then why are the attributes needed?

colombod: I prefer to have them if the y get serialized outside of the api. Ideally i'd stick them on the domain commands and events

jonsequitur: If that happens then the contract on the nested objects would be broken anyway.

For this reason I often make serialization-specific wrapper classes internal. The serialized output is the contract, and the class is an implementation detail that isn't intended to be used for other purposes.

WorkspaceServer.Tests/Kernel/KernelClientTests.cs
```diff
@@ -88,10 +91,45 @@ public async Task Kernel_client_surfaces_json_errors()
...
+            this.Assent(text, _configuration);
+        }
+
+        [Fact]
+        public async Task Kernel_client_surfaces_code_submission_Errors()
+        {
+            var kernel = new CompositeKernel()
+            {
+                new CSharpKernel(),
+                new FakeKernel("fake")
```
jonsequitur: Is the `FakeKernel` needed for this test?

WorkspaceServer.Tests/Kernel/KernelClientTests.cs
```diff
@@ -122,14 +160,7 @@ public async Task Kernel_can_pound_r_nuget_using_kernel_client()
...
             var text = reader.ReadToEnd();
-            var events = text.Split(new[] { Environment.NewLine }, StringSplitOptions.RemoveEmptyEntries)
-                .Select(JsonConvert.DeserializeObject<StreamKernelEvent>).ToList();
-
-            events.Should()
-                .Contain(e => e.EventType == nameof(NuGetPackageAdded));
-
-            events.Should()
-                .NotContain(e => e.EventType == nameof(CommandNotRecognized));
+            this.Assent(text, _configuration);
```
jonsequitur: This might be overly strict. Assent should be used for tests that are explicitly contract tests. This was (until now, and in terms of naming) a test of functionality. We should separate these so test failures are more informative and functional tests are resilient to normal refactoring.

PR #426, opened 2019-09-05
Pipeline fixes and script state access

jonsequitur:
This resolves an issue where the kernel's actual command handling was happening after pipeline execution returns, rather than as the last step in the pipeline.

Fixing this unblocks the ability to see `CSharpKernel.ScriptState` either before or after a submission is applied.

### DIFF
--- a/Microsoft.DotNet.Interactive.FSharp/FSharpKernel.fs
+++ b/Microsoft.DotNet.Interactive.FSharp/FSharpKernel.fs
@@ -32,7 +32,7 @@ type FSharpKernel() =
                 context.Publish(ReturnValueProduced(value, codeSubmission, formattedValues))
             | Ok(None) -> ()
             | Error(ex) -> context.OnError(ex)
-            context.Publish(CodeSubmissionEvaluated(codeSubmission))
+            context.Publish(CommandHandled(codeSubmission))
             context.Complete()
         }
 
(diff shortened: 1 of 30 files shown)


Microsoft.DotNet.Interactive.FSharp/FSharpKernel.fs
```diff
@@ -32,15 +32,13 @@ type FSharpKernel() =
                 context.Publish(ReturnValueProduced(value, codeSubmission, formattedValues))
             | Ok(None) -> ()
             | Error(ex) -> context.OnError(ex)
-            context.Publish(CodeSubmissionEvaluated(codeSubmission))
-            context.Complete()
+            context.Publish(CommandHandled(codeSubmission))
```
colombod: maybe we could use the Complete to actually publish the CommandHandled ?

Microsoft.DotNet.Interactive/KernelInvocationContext.cs
```diff
@@ -25,17 +23,10 @@ public class KernelInvocationContext : IDisposable
...
             _parentContext = parentContext;
             Command = command;
-            _invocation = command.InvokeAsync;
         }
 
         public IKernelCommand Command { get; }
 
-        public void Complete()
-        {
-            IsCompleted = true;
```
colombod: we might still need this part

PR #427, opened 2019-09-05
Publish interactive packages

akshita31:
We would like to publish the `Microsoft.Dotnet.Interactive` and `Microsoft.Dotnet.Interactive.Rendering` packages so that it is possible to write extensions to the kernel.

This PR:
1. Removes the dependency on MLS.Agent.Tools and publishes Microsoft.Dotnet.Interactive and Microsoft.DotNet.Interactive.Rendering nuget packages
2. Removes clockwise dependency from the MLS.Agent.Tools
2. For the Microsoft.Dotnet.Try.Markdown package it publishes the package with the MLS.Agent.Tools as a dll inside it so that we dont have to publish it separately as a nuget package


akshita31:
The latest commit https://github.com/dotnet/try/pull/427/commits/d9fd5ef02e59579e409a764d224d4704108ff0f8 adds a "beta" label for the library nuget packages and maintains the current behavior for the tool.

akshita31:
Many thanks to @brettfo 🥇

MLS.Agent.Tools/MLS.Agent.Tools.csproj
```diff
@@ -17,16 +17,17 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Clockwise" Version="1.0.261-beta" />
-    <PackageReference Include="Newtonsoft.Json" Version="12.0.02" />
-    <PackageReference Include="PocketLogger" Version="0.3.0">
+  <!-- Any new dependency added here must also be included in the Microsoft.DotNet.Try.Markdown, Microsoft.DotNet.Interactive and Microsoft.DotNet.Interactive.Rendering nuspec files -->
+    <PackageReference Include="Clockwise" Version="$(ClockwiseVersion)" />
```
rchande: I'm really not a fan of this--we're doing something that explicitly makes things weirder and harder in the future (I acknowledge it was not your idea).

rchande:
@jonsequitur Why don't we want to ship an MLS.Agent.Tools nuget package? The types within it are going to be public APIs regardless of what nuget package they ultimately appear in?

MLS.Agent.Tools/MLS.Agent.Tools.csproj
jonsequitur: We should actually look to remove the Clockwise dependency from here.

Microsoft.DotNet.Interactive.Rendering/Microsoft.DotNet.Interactive.Rendering.nuspec
```diff
@@ -0,0 +1,28 @@
...
+        $CommonMetadataElements$
+        <language>en-US</language>
+         <dependencies>
+            <group targetFramework=".NETStandard2.0">
+                <!-- Dependencies of Microsoft.Dotnet.Interactive.Rendering -->
+                <dependency id="Microsoft.AspNetCore.Html.Abstractions" version="$MicrosoftAspNetCoreHtmlAbstractionsVersion$" />
+                <dependency id="microsoft.csharp" version="$microsoftcsharpVersion$" />
+               
+                <!-- Dependencies of MLS.Agent.Tools -->
+                <dependency id="Clockwise" version="$ClockwiseVersion$" />
```
jonsequitur: This should not have a Clockwise dependency.

Microsoft.DotNet.Try.Markdown/Microsoft.DotNet.Try.Markdown.nuspec
```diff
@@ -0,0 +1,29 @@
...
+        <language>en-US</language>
+        <dependencies>
+            <group targetFramework=".NETStandard2.0">
+                <!-- Dependencies of Microsoft.DotNet.Try.Markdown -->
+                <dependency id="Microsoft.AspNetCore.Html.Abstractions" version="$MicrosoftAspNetCoreHtmlAbstractionsVersion$" />
+                <dependency id="system.commandline.experimental" version="$SystemCommandlineExperimentalVersion$" />
+                <dependency id="Markdig" version="$MarkdigVersion$" />
+
+                <!-- Dependencies of MLS.Agent.Tools -->
+                <dependency id="Clockwise" version="$ClockwiseVersion$" />
```
jonsequitur: This should not have a Clockwise dependency.

Microsoft.DotNet.Interactive.Rendering/Microsoft.DotNet.Interactive.Rendering.nuspec
akshita31: Moved Dotnet and CommandLine classes upto Workspace server hence removing the clockwise dependency here

Microsoft.DotNet.Try.Markdown/Microsoft.DotNet.Try.Markdown.nuspec
akshita31: Done

akshita31:
In the latest commit, I have removed the dependency of Microsoft.Dotnet.Interactive on MLS.Agent.Tools

WorkspaceServer/Kernel/CSharpKernel.cs
```diff
@@ -323,9 +324,10 @@ bool ShouldRebuild()
             }
         }
 
-        public async Task LoadExtensionsInDirectory(IDirectoryAccessor directory, KernelInvocationContext context)
+        public async Task LoadExtensionsInDirectory(DirectoryInfo directory, KernelInvocationContext context)
```
jonsequitur: Since this doesn't rely directly on private state, this is a good candidate for an extension method.

Is this code C# specific?

akshita31: This method is needed for implementing the interface `IExtensibleKernel`

PR #430, opened 2019-09-09
produce execute_input message when execute_request is handled

colombod:
address incomplete protocol implementation to improve experience with nteract components


captainsafia:
Nice! For future GitHub archaeologists, here is the related issue in the nteract repo: https://github.com/nteract/nteract/issues/4541

PR #431, opened 2019-09-09
Progress reporting for #r nuget

rchande:


WorkspaceServer.Tests/Kernel/CSharpKernelTests.cs
```diff
@@ -410,6 +410,16 @@ public async Task When_SubmitCode_command_adds_packages_to_csharp_kernel_then_Pa
                                .ToEnumerable()
                                .ToArray();
 
+            events
+                .First()
+                .Should()
+                .Match(e => e is DisplayedValueProduced && ((DisplayedValueProduced)e).Value.ToString().Contains("Attempting to install"));
```
colombod: probably here we should assert on a sequence of events

WorkspaceServer/Kernel/CSharpKernelExtensions.cs
```diff
@@ -68,22 +69,50 @@ public static CSharpKernel UseNugetDirective(this CSharpKernel kernel, INativeAs
                 {
                     Handler = async context =>
                     {
-                        var refs = await restoreContext.AddPackage(package.PackageName, package.PackageVersion);
+                        var message = $"Attempting to install package {package.PackageName}";
+                        if (!string.IsNullOrWhiteSpace(package.PackageVersion))
+                        {
+                            message += $", version {package.PackageVersion}";
+                        }
```
jonsequitur: Can we also get a notification about which actual version was installed in the case where the version was not specified?

WorkspaceServer/Kernel/CSharpKernelExtensions.cs
```diff
@@ -68,22 +69,50 @@ public static CSharpKernel UseNugetDirective(this CSharpKernel kernel, INativeAs
...
 
-                            kernel.AddMetadataReferences(refs);
+                            kernel.AddMetadataReferences(result.References);
+
+                            context.Publish(new DisplayedValueProduced($"Successfully added reference to package {package.PackageName}", context.Command));
+                            context.Publish(new NuGetPackageAdded(package));
+                        }
+                        else
+                        {
+                            context.Publish(new DisplayedValueProduced($"Failed to add reference to package {package.PackageName}", context.Command));
```
jonsequitur: When it fails, why does it fail, and is there additional information we can surface?

WorkspaceServer/Packaging/PackageRestoreContext.cs
```diff
@@ -4,12 +4,25 @@
...
 using Clockwise;
 using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.VisualBasic.Syntax;
 using MLS.Agent.Tools;
 
 namespace WorkspaceServer.Packaging
 {
     public class PackageRestoreContext
     {
+        public class AddReferenceResult
```
jonsequitur: Any reason for this to be a nested class?

WorkspaceServer/Kernel/CSharpKernelExtensions.cs
```diff
@@ -63,23 +64,52 @@ public static CSharpKernel UseNugetDirective(this CSharpKernel kernel, INativeAs
...
                 addPackage.Handler = async context =>
                 {
-                    var refs = await restoreContext.AddPackage(package.PackageName, package.PackageVersion);
+                    var message = $"Attempting to install package {package.PackageName}";
+                    if (!string.IsNullOrWhiteSpace(package.PackageVersion))
+                    {
+                        message += $", version {package.PackageVersion}";
+                    }
+
+                    var key = message;
```
jonsequitur: Is the redundant variable needed?

jonsequitur: Maybe call this `valueId` so it's clearer what its significance is.

WorkspaceServer/Kernel/CSharpKernelExtensions.cs
rchande: Will do this in a subsequent change

PR #433, opened 2019-09-10
%%time command, refactoring

jonsequitur:


Microsoft.DotNet.Interactive.Jupyter.Tests/ExecuteRequestHandlerTests.cs
```diff
@@ -48,23 +49,6 @@ public async Task handles_executeRequest()
             await handler.Handle(new JupyterRequestContext(_serverChannel, _ioPubChannel, request, _kernelStatus));
         }
 
-        [Fact]
-        public async Task sends_ExecuteInput_when_ExecuteRequest_is_handled()
```
colombod: why is this test removed?

PR #435, opened 2019-09-10
Make package logging more explicit

rchande:


MLS.Agent.Tools/AddPackageResult.cs
```diff
@@ -0,0 +1,53 @@
...
+            {
+                return output.Concat(error).Where(l => l.StartsWith("error:"));
+            }
+
+            private string GetInstalledVersion(IReadOnlyCollection<string> output)
+            {
+                foreach (var line in output)
+                {
+                    if (line.StartsWith("info : PackageReference for package"))
+                    {
```
brettfo: Maybe a quick check for negative values?  If the nuget output ever changes then this will crash.

rchande: Done.

MLS.Agent.Tools/AddPackageResult.cs
```diff
@@ -0,0 +1,59 @@
...
+                        // Successful installation will print a message like
+                        // @"info : PackageReference for package 'xunit' version '2.4.1' added to file 'E:\testdll\testdll.csproj'.";
+                        var quote1 = line.IndexOf('\'');
+                        if (quote1 == -1)
+                        {
+                            continue;
+                        }
+
+                        var quote2 = line.IndexOf('\'', quote1 + 1);
```
colombod: so it is 0 or 4 ```\``` char always?

rchande: Can't have ' in package name or version number...

MLS.Agent.Tools/Dotnet.cs
```diff
@@ -32,13 +33,13 @@ public Task<CommandLineResult> New(string templateName, string args = null, Budg
             return Execute($@"new ""{templateName}"" {args}", budget);
         }
 
-        public Task<CommandLineResult> AddPackage(string packageId, string version = null, Budget budget = null)
+        public async Task<AddPackageResult> AddPackage(string packageId, string version = null, Budget budget = null)
```
jonsequitur: I think this can also just be an extension method since it doesn't rely on any non-public state or APIs of the `DotNet` class.

PR #440, opened 2019-09-11
Use msbuild to get the path of the nuget packages

akshita31:
We needed a way to discover KernelExtensions that are part of nuget packages that do not have a "lib" folder.

This PR uses msbuild for doing so. It adds targets which does the following things:
1. Add `GeneratePathProperty` to the PackageReference property group. When this is set then in the obj folder, you will be able to see a nuget.g.props file which has a property that contains the path to the nuget package, for example, `PkgFluentAssertions`

image

2. The target `ComputePackageRoots` reads this property for all the nuget packages and stores the paths in item `AddedNuGetPackage`.

3. The third target prints the values from `AddedNuGetPackage` target into a file `*.nuget.paths`
image

Then in the package restore context we read this path and try to load extensions if any are present.


akshita31:
cc @KevinRansom

WorkspaceServer.Tests/PackageRestoreContextTests.cs
```diff
@@ -28,5 +29,14 @@ public async Task Returns_failure_if_package_installation_fails()
...
             result.DetailedErrors.Should().NotBeEmpty();
         }
+
+        [Fact]
+        public async Task Can_get_path_to_nuget_package()
+        {
+            PackageRestoreContext packageRestoreContext = new PackageRestoreContext();
+            await packageRestoreContext.AddPackage("fluentAssertions", "5.7.0");
+            var path = await packageRestoreContext.GetDirectoryForPackage("fluentassertions");
+            path.FullName.Should().EndWith("fluentassertions" + Path.DirectorySeparatorChar + "5.7.0");
```
jonsequitur: A more thorough test would be to check that the file exists.

Microsoft.DotNet.Interactive/Commands/LoadExtensionInDirectory.cs
```diff
@@ -0,0 +1,21 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using MLS.Agent.Tools;
+using Newtonsoft.Json;
+
+namespace Microsoft.DotNet.Interactive.Commands
+{
+    public class LoadExtensionsInDirectory : KernelCommandBase
```
jonsequitur: File name doesn't match class name.

akshita31: Done. Thanks

PR #449, opened 2019-09-16
Build on rc1

rchande:


rchande:
Note that this won't work in CI due to https://github.com/aspnet/AspNetCore/issues/14040, but i think that will be resolved by the GA of .NET Core 3

PR #463, opened 2019-09-19
WIP: Add a sample kernel extension

akshita31:


jonsequitur:
This kind of functionality should work via existing display functionality rather than by introducing a new helper method. I'd be interested to include `IObservable` display support out of the box, though, rather than via an extension, since it's very generalized.

PR #467, opened 2019-09-19
Add writefile magic command

akshita31:


akshita31:
I will close this PR for now and come back to it once I have more time to complete it

PR #476, opened 2019-09-23
Add Telemetry

TIHan:
This adds Telemetry to dotnet-try CLI, similar to that of dotnet CLI. 

To opt-out of telemetry, set the environment variable, "DOTNET_TRY_CLI_TELEMETRY_OPTOUT", to "true", "1", or "yes".

This is almost ready.


MLS.Agent.Tests/TelemetryTests.cs
```diff
@@ -73,21 +73,26 @@ public void Dispose()
         }
 
         [Fact]
-        public async Task TelemetryCommandIsValid()
+        public async Task Jupyter_standalone_command_is_valid()
```
jonsequitur: "is valid" is pretty vague. The test is actually verifying that the command, when invoked without arguments, sends telemetry.

TIHan: Yea, it is vague. I should be more descriptive, "is_valid" => "sends_telemetry".

MLS.Agent/Telemetry/Configurer/IFirstTimeUseNoticeSentinel.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace MLS.Agent.Telemetry.Configurer
 {
-    internal interface IFirstTimeUseNoticeSentinel : IDisposable
+    public interface IFirstTimeUseNoticeSentinel : IDisposable
```
jonsequitur: From the usages I see, it doesn't look like this needs to be `IDisposable`.

PR #480, opened 2019-09-25
create protocol adapter

colombod:
The goal is to strengthen the transformation between domain entities and jupyter protocol
- addresses also #437


Microsoft.DotNet.Interactive.Jupyter/IPubSubChannel.cs
```diff
@@ -0,0 +1,12 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.DotNet.Interactive.Jupyter.Protocol;
+
+namespace Microsoft.DotNet.Interactive.Jupyter
+{
+    public interface IPubSubChannel
+    {
+        bool Send(JupyterPubSubContent content, Message request, string ident = null);
+    }
```
jonsequitur: Why are content and request different?

jonsequitur: Why the bool return type? What does it signify? Also, should this be async?

jonsequitur: What's the use case and expectation for ident being null?

Microsoft.DotNet.Interactive.Jupyter/ReplyChannel.cs
```diff
@@ -0,0 +1,38 @@
...
+        }
+        public bool Send(JupyterReplyContent content, Message request)
+        {
+            var reply = Message.CreateReply(content, request);
+            return _sender.Send(reply);
+        }
+    }
+
+    public class PubSubChannel : IPubSubChannel
+    {
```
jonsequitur: Separate file please.

Microsoft.DotNet.Interactive.Jupyter/IPubSubChannel.cs
colombod: not always needed to generate the topic

PR #482, opened 2019-09-26
Consolidate interactive tests

KevinRansom:
First attempt at rationalizing F# and C# interactive test cases.

Simplify and refactor: CSharpKernelTests/FSharpKernelTests/CSharpKernelTestBase/FSharpKernelTestBase/KernelTestBase into

LanguageKernelTests/LanguageKernelTestBase

Here is an example test case with C# and F# source variants:
https://github.com/dotnet/try/compare/master...KevinRansom:kr?expand=1#diff-d8de318123580f112042b15d5a087f4dR55

**I really wish this was in F# the amount of ceremony in C# is unbelievable
In fact I may port this whole thing to F# just to show you.**

````
        [Theory]
        [InlineData(Language.FSharp)]
        [InlineData(Language.CSharp)]
        public async Task it_remembers_state_between_submissions(Language language)
        {
            string[] source = null;
            switch (language)
            {
                case Language.FSharp:
                    string[] fsLines =
                    {
                        "let add x y = x + y",
                        "add 2 3"
                    };
                    source = fsLines;
                    break;
                case Language.CSharp:
                    string[] csLines =
                    {
                        "int Multiply(int x, int y) { return x + y; };",
                        "Multiply(2, 3);"
                    };
                    source = csLines;
                    break;
            }

            var kernel = CreateKernel(language);

            await SubmitLines(kernel, source);

            KernelEvents.ValuesOnly()
                .OfType<ReturnValueProduced>()
                .Last()
                .Value
                .Should()
                .Be(5);
        }
````

### CODE REVIEW

File: WorkspaceServer.Tests/Kernel/LanguageKernelTestBase.cs [hunk 1]
```diff
@@ -0,0 +1,86 @@
...
+            return kernel;
+        }
+
+        protected KernelBase CreateKernel()
+        {
+            return CreateKernel(Language.CSharp);
+        }
+
+        public async Task SubmitLines(KernelBase kernel, string[] lines)
+        {
```
jonsequitur: The naming is a bit misleading since multi-line submissions are very common.

jonsequitur:
Generally, agreed, though a switch expression would make this example a little more concise:

```cs
        [Theory]
        [InlineData(Language.FSharp)]
        [InlineData(Language.CSharp)]
        public async Task it_remembers_state_between_submissions(Language language)
        {
            var source = language switch
            {
                Language.FSharp => new[]
                {
                    "let add x y = x + y",
                    "add 2 3"
                },
                Language.CSharp => new[]
                {
                    "int Multiply(int x, int y) { return x + y; };",
                    "Multiply(2, 3);"
                }
            };

            var kernel = CreateKernel(language);

            await SubmitLines(kernel, source);

            KernelEvents.ValuesOnly()
                        .OfType<ReturnValueProduced>()
                        .Last()
                        .Value
                        .Should()
                        .Be(5);
        }
```

jonsequitur:
It would be helpful for reviewers to spot possible errors if you do the renames in a prior commit to the actual changes.

WorkspaceServer.Tests/Kernel/LanguageKernelTestBase.cs
```diff
@@ -0,0 +1,86 @@
...
+            var kernel = CreateLanguageKernel(language).LogEventsToPocketLogger();
+
+            DisposeAfterTest(
+                kernel.KernelEvents.Timestamp().Subscribe(KernelEvents.Add));
+
+            return kernel;
+        }
+
+        protected KernelBase CreateKernel()
```
brettfo: I'd feel better if this function was removed.  Tests that call this are really C#-specific tests so I'd like them to explicitly call `CreateKernel(CSharp)` _or_ group them in a new class with `IKernel CreateKernel() => base.CreateKernel(CSharp);` just so it's explicit..

WorkspaceServer.Tests/Kernel/LanguageKernelTestBase.cs
```diff
@@ -0,0 +1,86 @@
...
+        {
+            return CreateKernel(Language.CSharp);
+        }
+
+        public async Task SubmitLines(KernelBase kernel, string[] lines)
```
brettfo: `params string[] lines` might be easier to consume, but ultimately I don't care too much.

brettfo:
General request: To make the test cases easier to parse, could the switch statements/expressions be replaced with this?

``` C#
// I'm not married to the name of this function, but I think the use of `T` will make it easy to in the future return more than just strings.
public T GetElementForLanguage<T>(Language language, params (Language lang, T value)[] values)
{
    // The use of `Single()` ensures that the test will fail if a value was not given for a specific language.
    return values.Single(val => val.lang == language).value;
}
```

That way the call sites will be a little more concise:

``` C#
var code = GetElementForLanguage(language, (Language.CSharp, "using System;"),
                                           (Language.FSharp, "open System"));
```

KevinRansom:
> Generally, agreed, though a switch expression would make this example a little more concise:

@jonsequitur, I tried it, unfortunately an array initializer is not an expression that can be returned from a switch.

KevinRansom:
> Generally, agreed, though a switch expression would make this example a little more concise:

@jonsequitur, I tried it, unfortunately an array initializer is not an expression that can be returned from a switch expression.

KevinRansom:
> Generally, agreed, though a switch expression would make this example a little more concise:
> 
> @jonsequitur, I tried it, unfortunately an array initializer is not an expression that can be returned from a switch expression.

Oh the trick is to use  new[].  Nice.

KevinRansom:
I have added 4 different ways of doing a test based on the feedback, and other ideas.  let me know what you like best.

WorkspaceServer.Tests/Kernel/LanguageKernelTests.cs
```diff
@@ -45,6 +48,141 @@ public async Task it_returns_the_result_of_a_non_null_expression()
...
+        {
+            // The use of `Single()` ensures that the test will fail if a value was not given for a specific language.
+            return values.Single(val => val.lang == language).value;
+        }
+        
+        //Option 3: GetElementForLanguage iter 1
+        [Theory]
+        [InlineData(Language.FSharp)]
+        [InlineData(Language.CSharp)]
+        public async Task it_remembers_state_between_submissions_option3(Language language)
```
brettfo: I'm going to cast my vote for this style of test case because:

1.  The code to select different values is short and easy to read.  I find the `switch` expression difficult because it's so verbose and requires a `default` case that must explicitly throw, while the helper method here handles that for you.
1.  Placing code in `[InlineData]` works for compile-time constants, but won't be usable for expressions.  We don't have a test that does this currently, but I can easily see a test that needs to call `GetElementForLanguage()` where the item returned is a `Type` or `Exception` or something that can't be placed in an attribute.
1.  Finally, I like that the code variants are declared exactly where and when they're used instead of at the top of the function where I have to trace a variable through.

WorkspaceServer.Tests/Kernel/LanguageKernelTestBase.cs
jonsequitur: I think `submissions` is fine.

---
In reply to: [328715808](https://github.com/dotnet/try/pull/482#discussion_r328715808) [](ancestors = 328715808)

WorkspaceServer.Tests/Kernel/LanguageKernelTestBase.cs
```diff
@@ -0,0 +1,97 @@
...
+            {
+                var cmd = new SubmitCode(line, targetKernelName: targetKernelName, submissionType: submissionType);
+                await kernel.SendAsync(cmd);
+                cmds.Add(cmd);
+            }
+            return cmds.ToArray();
+        }
+
+        public async Task<SubmitCode> SubmitSource(KernelBase kernel, string line, string targetKernelName = null, SubmissionType submissionType = SubmissionType.Run)
+        {
```
jonsequitur: To avoid vocabulary creep, `SubmitCode` would be better.

WorkspaceServer.Tests/Kernel/LanguageKernelTestBase.cs
```diff
@@ -0,0 +1,97 @@
...
+            foreach (string line in lines)
+            {
+                var cmd = new SubmitCode(line, targetKernelName: targetKernelName, submissionType: submissionType);
+                await kernel.SendAsync(cmd);
+                cmds.Add(cmd);
+            }
+            return cmds.ToArray();
+        }
+
+        public async Task<SubmitCode> SubmitSource(KernelBase kernel, string line, string targetKernelName = null, SubmissionType submissionType = SubmissionType.Run)
```
jonsequitur: I suspect that anywhere we're passing `targetKernelName` it's actually unnecessary.

KevinRansom:
@colombod,

SubmitSource is an overloaded method, so you can pass in a line or an array of lines.  If we determine that we need an array of arrays then we can always add that overload when the need presents itself.

Kevin

WorkspaceServer.Tests/Kernel/LanguageKernelTestBase.cs
KevinRansom: removed

WorkspaceServer.Tests/Kernel/LanguageKernelTestBase.cs
KevinRansom: Sure

WorkspaceServer.Tests/Kernel/LanguageKernelTests.cs
```diff
@@ -0,0 +1,1117 @@
...
+            {
+                Language.FSharp => "let a =",
+                Language.CSharp => "var a ="
+            };
+
+            await SubmitCode(kernel, source, submissionType: SubmissionType.Diagnose);
+
+            KernelEvents
+                .ValuesOnly()
+                .Single(e => e is IncompleteCodeSubmissionReceived);
```
jonsequitur: >        [](start = 9, length = 7)

`.Should().ContainSingle()`

WorkspaceServer.Tests/Kernel/LanguageKernelTests.cs
```diff
@@ -0,0 +1,1117 @@
...
+        [InlineData(Language.CSharp)]
+        //[InlineData(Language.FSharp)]                     // Todo: FSI - returns a value
+        public async Task it_can_analyze_complete_stdio_submissions(Language language)
+        {
+            var kernel = CreateKernel(language);
+
+            var source = language switch
+            {
+                Language.FSharp => "Console.WriteLine(\"Hello\")",
+                Language.CSharp => "Console.WriteLine(\"Hello\")"
```
jonsequitur: In this case the test will pass for the wrong reason because this is not a complete C# statement unless you add a semicolon to the end.

WorkspaceServer.Tests/Kernel/LanguageKernelTests.cs
```diff
@@ -0,0 +1,1117 @@
...
+            {
+                Language.FSharp => new[] {
+$"#r @\"{dllPath}\"",
+@"
+open Newtonsoft.Json
+let json = JsonConvert.SerializeObject( struct {| value = ""hello"" |} )
+json
+"},
+                Language.CSharp => new[] {
+$"#r @\"{dllPath}\"",
```
jonsequitur: >$"#r @\"{dllPath}\"" [](start = 0, length = 20)

I believe this works if you remove the `@`. The issue is not the quoted path.

KevinRansom: Ahh … nice.

WorkspaceServer.Tests/Kernel/LanguageKernelTests.cs
KevinRansom: @jonsequitur 
Console.WriteLine("Hello, World") is a complete C# expression, if it was incomplete this would fail:

````
            KernelEvents
                .Should()
                .Contain(e => e.Value is CompleteCodeSubmissionReceived);
````
This is the notebook with a C# kernel note no ';'
image

PR #490, opened 2019-10-01
preserve ordering of commands when splitting a submission

jonsequitur:


WorkspaceServer.Tests/Kernel/CSharpKernelTests.cs
```diff
@@ -383,15 +383,21 @@ public async Task it_returns_completion_list_for_types_imported_at_runtime()
         [Fact]
         public async Task When_SubmitCode_command_adds_packages_to_csharp_kernel_then_the_submission_is_not_passed_to_csharpScript()
         {
+            var cSharpKernel = new CSharpKernel();
```
KevinRansom: Any chance you could go back to:
new CSharpKernel().UseNugetDirective()

And place the 
 using var events = kernel.KernelEvents.ToSubscribedList();

After all it will be eventually it will be:
new Kernel(language);  after all we have multiple languages :-)

PR #493, opened 2019-10-02
Improve command serialization for streaming kernel commands

colombod:


WorkspaceServer.Tests/Kernel/StreamKernelCommandTests.cs
```diff
@@ -0,0 +1,44 @@
...
+        {
+            var command = new SubmitCode("display(x); display(x + 1); display(x + 2);");
+            var streamCommand =  new StreamKernelCommand
+            {
+                Id = 123,
+                CommandType = command.GetType().Name,
+                Command = command
+            };
+
+            var serializedCommand = JsonConvert.SerializeObject(streamCommand, _jsonSerializerSettings);
```
jonsequitur: >   var serializedCommand = JsonConvert.SerializeObject(streamCommand, _jsonSerializerSettings); [](start = 9, length = 95)

This test is more valid if the serializer settings or serialization method is in the product code, not in a test.

colombod: done!

PR #494, opened 2019-10-02
standar io messages will have specific value produced events

colombod:
stdout -> StandardOutputValueProduced
sterror -> StandardErrorValueProduced

those events will be mapped to Stream jupyter messages

### DIFF
--- a/Microsoft.DotNet.Interactive.Jupyter/ExecuteRequestHandler.cs
+++ b/Microsoft.DotNet.Interactive.Jupyter/ExecuteRequestHandler.cs
@@ -119,13 +119,13 @@ namespace Microsoft.DotNet.Interactive.Jupyter
                         transient: transient,
                         data: formattedValues));
                     dataMessage.Enqueue(Stream.StdOut(
-                        GetPlainTextValueOrDefault(formattedValues, value?.ToString()))
+                        GetPlainTextValueOrDefault(formattedValues, value?.ToString() ?? string.Empty))
                     );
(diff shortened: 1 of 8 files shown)


Microsoft.DotNet.Interactive.Jupyter/ExecuteRequestHandler.cs
```diff
@@ -106,30 +95,64 @@ public async Task Handle(JupyterRequestContext context)
 
             CreateDefaultFormattedValueIfEmpty(formattedValues, value);
 
-            PubSubMessage executeResultData;
+            Queue<PubSubMessage> dataMessage = new Queue<PubSubMessage>();
```
jonsequitur: Feel free to use `var`.

colombod: will do!

WorkspaceServer/Kernel/CSharpKernel.cs
```diff
@@ -246,6 +248,30 @@ public Task<bool> IsCompleteSubmissionAsync(string code)
                     output,
                     command,
                     formattedValues));
```
jonsequitur: We should stop publishing this event here. The Jupyter layer should decide how to handle StandardOutputValueProduced.

colombod: makes sense

PR #511, opened 2019-10-09
Support #r @" " and #r """ """ strings

KevinRansom:
Disable response file support for kernel directives.  This ensures that @"" and """ """ are passed through to interactive.

Re-enable the tests for this.


Microsoft.DotNet.Interactive/Extensions/CommandLineBuilderExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+
+using System;
+using System.CommandLine;
+using System.CommandLine.Builder;
+
+namespace Microsoft.DotNet.Interactive.Extensions
+{
+    public static class CommandLineBuilderExtensions
+    {
+        public static CommandLineBuilder ResponseFile(
```
jonsequitur: This method already exists `System.CommandLine`:

https://github.com/dotnet/command-line-api/blob/0fd4b6381f62993c16414b00fcb4804302d04890/src/System.CommandLine/Builder/CommandLineBuilderExtensions.cs#L100-L106

KevinRansom: Jon … it's a property.  The helper method sets the property.

jonsequitur: The method in `System.CommandLine` is also setting the property and also returning the `CommandLineBuilder` instance. The implementation is identical (other than the null check).

KevinRansom:
@colombod can we get this merged pls.

PR #512, opened 2019-10-09
Fix output

KevinRansom:
Console and output from interactive to Jupyter, had a couple of issues.

1.  It enqueued StandardOutputValueProduced  twice, doubling up the output.

2. It didn't handle protocol stream correctly.  So it only actually produced the first line.

Repro:
Console.WriteLine a couple of lines, and see only one produced.  Either C# or F#.

image

With Fix:
image

### DIFF
--- a/Microsoft.DotNet.Interactive.Jupyter/ZMQ/Message.cs
+++ b/Microsoft.DotNet.Interactive.Jupyter/ZMQ/Message.cs
@@ -144,12 +144,17 @@ namespace Microsoft.DotNet.Interactive.Jupyter.ZMQ
                     break;
 
                 case nameof(DisplayData):
+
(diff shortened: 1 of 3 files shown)


Microsoft.DotNet.Interactive.Jupyter/ZMQ/Message.cs
```diff
@@ -152,6 +153,10 @@ public class Message
                     encodedTopic = Encoding.Unicode.GetBytes("execute_result");
                     break;
 
+                case nameof(Stream):
```
colombod: look at code in #513

colombod:
This Pr is replaced by PR #514

colombod:
this PR is addressed by #514

PR #513, opened 2019-10-09
generates topics for stream messages

colombod:


Microsoft.DotNet.Interactive.Jupyter/ZMQ/Message.cs
```diff
@@ -156,6 +156,16 @@ public class Message
                     encodedTopic = null;
                     break;
 
+                case nameof(Stream):
+                    {
+                        if (!(content is Stream stream))
```
KevinRansom: @colombod , I am pretty sure that this type test is redundant.
After all the name came from GetType().

However, C# supports type case so probably better way to fix this is just to switch on type.

Kevin

PR #516, opened 2019-10-09
terminal formatter and mime type support

colombod:
Provide vt100 formatting in rendering

### DIFF
--- a/Microsoft.DotNet.Interactive.Jupyter/ExecuteRequestHandler.cs
+++ b/Microsoft.DotNet.Interactive.Jupyter/ExecuteRequestHandler.cs
@@ -41,7 +41,7 @@ namespace Microsoft.DotNet.Interactive.Jupyter
         }
 
         protected override void OnKernelEventReceived(
-            IKernelEvent @event, 
+            IKernelEvent @event,
             JupyterRequestContext context)
         {
             switch (@event)
@@ -53,7 +53,7 @@ namespace Microsoft.DotNet.Interactive.Jupyter
                     OnCommandHandled(context.JupyterMessageSender);
                     break;
                 case CommandFailed commandFailed:
-                    OnCommandFailed(commandFailed,  context.JupyterMessageSender);
+                    OnCommandFailed(commandFailed, context.JupyterMessageSender);
                     break;
             }
         }
@@ -82,7 +82,7 @@ namespace Microsoft.DotNet.Interactive.Jupyter
                     break;
             }
 
-            var errorContent = new Error (
+            var errorContent = new Error(
                 eName: "Unhandled Exception",
                 eValue: commandFailed.Message,
                 traceback: traceBack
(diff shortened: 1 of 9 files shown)


Microsoft.DotNet.Interactive.Jupyter/ExecuteRequestHandler.cs
```diff
@@ -78,11 +78,11 @@ public async Task Handle(JupyterRequestContext context)
                 default:
                     traceBack.Add("Unhandled Exception");
```
jonsequitur: Does the output look correct if the traceback is added as a single string with line breaks?

Microsoft.DotNet.Interactive.Rendering/TerminalFormatter.cs
```diff
@@ -0,0 +1,31 @@
...
+                    typeof(bool)
+                });
+
+            return (ITypeFormatter)genericCreateForAllMembers.Invoke(null, new object[]
+            {
+                includeInternals
+            });
+        }
+
+        public const string MimeType = "text/vt100";
```
jonsequitur: This is probably not the correct name for this mime type, since VT-100 is really a subset of the [ANSI escape sequences](https://en.wikipedia.org/wiki/ANSI_escape_code) that might be used.

Microsoft.DotNet.Interactive.Rendering/TerminalFormatter{T}.cs
```diff
@@ -0,0 +1,85 @@
...
+    public class TerminalFormatter<T> : TypeFormatter<T>
+    {
+        private readonly Action<T, TextWriter> _format;
+
+        public TerminalFormatter(Action<T, TextWriter> format)
+        {
+            _format = format;
+        }
+
+        public override void Format(T instance, TextWriter writer)
```
jonsequitur: Null check.

Microsoft.DotNet.Interactive.Rendering/TerminalFormatter{T}.cs
```diff
@@ -0,0 +1,85 @@
...
+using System.CommandLine.Rendering.Views;
+using System.IO;
+
+namespace Microsoft.DotNet.Interactive.Rendering
+{
+    public class TerminalFormatter<T> : TypeFormatter<T>
+    {
+        private readonly Action<T, TextWriter> _format;
+
+        public TerminalFormatter(Action<T, TextWriter> format)
```
jonsequitur: Null check.

Microsoft.DotNet.Interactive.Rendering/TerminalFormatter{T}.cs
```diff
@@ -0,0 +1,85 @@
...
+                return CreateForSequence(includeInternals);
+            }
+
+            return CreateForObject(includeInternals);
+        }
+
+        private static ITypeFormatter<T> CreateForObject(bool includeInternals)
+        {
+            var members = typeof(T).GetAllMembers(includeInternals)
+                .GetMemberAccessors<T>();
```
jonsequitur: A better approach here would be to use the `Destructurer` class, which is what the `HtmlFormatter` uses to build up the table.

PR #518, opened 2019-10-09
Real Instrumentation Key for Telemetry + plus minor fixes

TIHan:
This adds the real instrumentation key used for telemetry.

Also has some minor fixes on how we log optionals.


TIHan:
I can't merge this as I'm not authorized.

PR #519, opened 2019-10-09
handle compilation failures in the F# kernel

brettfo:
Includes bringing more language agnostic tests in line.

### DIFF
--- a/Microsoft.DotNet.Interactive.FSharp/FSharpKernel.fs
+++ b/Microsoft.DotNet.Interactive.FSharp/FSharpKernel.fs
@@ -6,6 +6,7 @@ namespace Microsoft.DotNet.Interactive.FSharp
 open System.Collections.Generic
 open System.IO
 open System.Threading.Tasks
+open FSharp.Compiler.Interactive.Shell
 open FSharp.Compiler.Scripting
 open Microsoft.DotNet.Interactive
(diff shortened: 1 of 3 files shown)


Microsoft.DotNet.Interactive.FSharp/FSharpKernel.fs
```diff
@@ -45,15 +46,21 @@ type FSharpKernel() as this =
...
-                context.Publish(ReturnValueProduced(value, codeSubmission, formattedValues))
-            | Ok(None) -> ()
+            | Ok(result) ->
+                match result with
+                | Some(value) ->
+                    let value = value.ReflectionValue
+                    let formattedValues = FormattedValue.FromObject(value)
+                    context.Publish(ReturnValueProduced(value, codeSubmission, formattedValues))
+                | None -> ()
+                context.Complete()
```
KevinRansom: @brettfo, is context.Complete() not needed when we match | Error?

jonsequitur: It shouldn't be. This part of the API is hard to understand, though. Ideally (in a future, better version of this API) the context is completed automatically when the handler has exited and this method call should become implicit or unnecessary and go away. The more interesting case is when you want the event channel to remain alive after the handler has exited, e.g. publishing or streaming data back to the client.

PR #521, opened 2019-10-11
Improve F# integration

KevinRansom:
Open a couple more default namespaces.
Add a couple more referenced assemblies


brettfo:
Adding namespaces has changed the errors returned in some of the tests, so you'll either have to update the tests or disallow namespace/type suggestions.

KevinRansom:
I'll update the tests, I like the suggestions.

PR #545, opened 2019-10-22
load dependencies at extension load time

colombod:
loading nuget packages only load references to script options.  Extension loading requires assemblies to be available via the AssemblyLoadContext

### DIFF
--- a/Microsoft.DotNet.Interactive/IExtensibleKernel.cs
+++ b/Microsoft.DotNet.Interactive/IExtensibleKernel.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Collections.Generic;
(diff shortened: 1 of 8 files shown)


Microsoft.DotNet.Interactive/IExtensibleKernel.cs
```diff
@@ -2,13 +2,12 @@
...
 using MLS.Agent.Tools;
-using System.IO;
 using System.Threading.Tasks;
 
 namespace Microsoft.DotNet.Interactive
 {
     public interface IExtensibleKernel
     {
-        public Task LoadExtensionsInDirectory(IDirectoryAccessor directory, KernelInvocationContext invocationContext);
+        Task LoadExtensionsInDirectory(IDirectoryAccessor directory, KernelInvocationContext invocationContext);
```
jonsequitur: Could we rename to `LoadExtensionsFromDirectory`?

colombod: done!

PR #546, opened 2019-10-22
enable additional F# tests

brettfo:
The bug where F# `let x = ...` returns a value was fixed sometime recently, so this PR simply turns on some F# tests.  The test about statements had an incorrect comment for it's C# snippet, so that was updated to be correct (n.b., that there is another C# test that's `var x = ...` so we're not losing any test coverage.)


brettfo:
At this point the only gap between C# and F# functionality for the kernels is completion.

Issue #558, opened 2019-10-25
Implicitly ignore the result of `display`


cartermp:
Currently, `display` returns a value that is evaluated by F#:

```fsharp
let names = [ "phillip"; "will"; "diego" ]
display names
```

![image](https://user-images.githubusercontent.com/6309070/67534422-e46c6080-f682-11e9-9fa4-df7173755684.png)

To not display that bottom part, F# needs to explicitly `ignore` it:

```fsharp
let names = [ "phillip"; "will"; "diego" ]
display names |> ignore
```
![image](https://user-images.githubusercontent.com/6309070/67534445-06fe7980-f683-11e9-8de5-bc17a0f6f75e.png)

Talking with @colombod, it doesn't seem to be particularly useful for either C# or F# to have that evaluated value. I propose that we implicitly ignore it, since any updates to the display won't be by manipulating the object and then evaluating it; they'll come from calling the `Update` method.

PR #576, opened 2019-10-30
do not produce ExecuteResult if the ReturnValueProduced events raw value is of type DisplayedValue

colombod:
Addressing issue #558 when the cell evaluates to a `DisplayedValue` the object should not be printed

### DIFF
--- a/WorkspaceServer.Tests/Kernel/ObservableExtensions.cs
+++ b/WorkspaceServer.Tests/Kernel/ObservableExtensions.cs
@@ -54,12 +54,17 @@ namespace WorkspaceServer.Tests.Kernel
                         termination.OnNext(Unit.Default);
                     }
                 })
-                .Timeout(timeout ?? 10.Minutes());
+                .Timeout(timeout ?? 1.Seconds());
         }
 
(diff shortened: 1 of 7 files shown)


WorkspaceServer.Tests/Kernel/ObservableExtensions.cs
```diff
@@ -12,43 +12,62 @@ namespace WorkspaceServer.Tests.Kernel
...
+            var stream =  source.TakeUntil(termination)
                 .Do(e =>
                 {
-                    if (e["eventType"].Value<string>() == nameof(CommandParseFailure))
+                    if (e["eventType"].Value<string>() == typeof(T).Name)
                     {
                         termination.OnNext(Unit.Default);
                     }
                 });
+            if (timeout != null)
```
jonsequitur: Since these are test helpers, it's simpler to not have to specify a timeout in all of the tests, and if `null` is passed, pass a default here. There's no valid use case for not having a timeout, except...

An additional nice-to-have there is to not pass the timeout if the debugger is attached.

PR #587, opened 2019-11-01
F# github sample

KevinRansom:
Address octokit assembly load issue.

Port github sample to f#


Microsoft.DotNet.Interactive.FSharp/FSharpKernel.fs
```diff
@@ -25,14 +25,6 @@ type FSharpKernel() as this =
     do Event.add resolvedAssemblies.Add script.AssemblyReferenceAdded
     do base.AddDisposable(script)
 
-    let handleAssemblyReferenceAdded path context =
```
colombod: this is not the real fix, i'd rather address the fact that `KernelExtensionLoader.LoadFromAssembly` should return immediately if the `kernel` parameter is no implementing the `IExtensibleKernel` inerface

colombod: To be complete this part of code is not consistent as is loading extensions if a reference is added. The `CSharpKernel` behaviour is quite different as that should be done when NuGet packages are loaded or if `#extend` is triggered. It is possible to load extension of the Kernel without loading assemblies into the engine itself.

KevinRansom: @colombod, chatting with @jonsequitur we think we probably don't want to search for dependencies every 
````#r "somassembly.dll".````

What we think we will do is raise a package root notification for every nuget package and dependency discovered, and that will then probe for kernel extensions.

@brettfo is planning on adding that shortly.  @jonsequitur is updating the #r "nuget: to get that information out of the c# implementation.

I can surely make a change the change you suggest however I expect extension  loading to be triggered differently.

Kevin

colombod: We should not do that on #r or assembly loading but only on #extend or nuget load. With the change I am proposing you will not get errors on loading oktokit

KevinRansom: @colombod,
however, the above event is fired whenever fsi #r's a dll.  There is clearly no reason at all for us to even initiate the KernelExtensionLoader at this point.  Effectively, we will always invoke the event to be ignored, so we may as well not invoke it at all, unless there is some benefit to us firing the event.

We should still generate and propagate and assembly referenced event so the kernel can choose to deal with it as it will.

KevinRansom: One other thing ... the KernelExtension code should be able to load the assembly fine, since it's a valid loadable assembly, albeit, unsigned.  I can and will submit a pr that will fix the load, however, before I do I want to chat with the coreclr guys to see if the failure is "by design", I think it is a bug.

KevinRansom:
@jonsequitur, @colombod.  I am not authorized to merge, so if we are ready, would one of you merge it please.

PR #591, opened 2019-11-02
display F# package installation status

brettfo:
Consume events added in dotnet/fsharp#7796 to display package installation status.

image

### DIFF
--- a/WorkspaceServer.Tests/Kernel/LanguageKernelTests.cs
+++ b/WorkspaceServer.Tests/Kernel/LanguageKernelTests.cs
@@ -11,6 +11,7 @@ using FluentAssertions.Extensions;
 using Microsoft.DotNet.Interactive;
 using Microsoft.DotNet.Interactive.Commands;
 using Microsoft.DotNet.Interactive.Events;
+using Microsoft.DotNet.Interactive.FSharp;
 using Microsoft.DotNet.Interactive.Tests;
 using MLS.Agent;
 using MLS.Agent.Tools;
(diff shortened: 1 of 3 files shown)


WorkspaceServer.Tests/Kernel/LanguageKernelTests.cs
```diff
@@ -943,28 +944,33 @@ public async Task When_SubmitCode_command_adds_packages_to_csharp_kernel_then_th
...
+            var command = new SubmitCode(code);
 
             var result = await kernel.SendAsync(command);
 
             using var events = result.KernelEvents.ToSubscribedList();
 
             events
-                .First()
                 .Should()
-                .Match(e => e is DisplayedValueProduced && ((DisplayedValueProduced)e).Value.ToString().Contains("Installing"));
```
brettfo: Since the F# kernel handles `#r "nuget:..."` natively, `DisplayedValueProduced` won't be the first event, but it _does_ eventually appear.

WorkspaceServer.Tests/Kernel/LanguageKernelTests.cs
```diff
@@ -943,28 +944,33 @@ public async Task When_SubmitCode_command_adds_packages_to_csharp_kernel_then_th
...
 
-        [Fact]
-        public async Task When_SubmitCode_command_adds_packages_to_csharp_kernel_then_PackageAdded_event_is_raised()
+        [Theory]
+        [InlineData(Language.CSharp, "Microsoft.Extensions.Logging.ILogger logger = null;")]
+        [InlineData(Language.FSharp, "let logger: Microsoft.Extensions.Logging.ILogger = null")]
+        public async Task When_SubmitCode_command_adds_packages_to_kernel_then_PackageAdded_event_is_raised(Language language, string expression)
         {
-            var kernel = new CompositeKernel
+            IKernel kernel = language switch
```
colombod: Can use var

brettfo: Can't because the expression is inferred to be of type `CSharpKernel` which makes the F# line fail type inference.  If I kept var I would have to add `as IKernel` after the C# one so I instead opted to type up front.

PR #600, opened 2019-11-06
Csi native

KevinRansom:
Update @jonsequitur 's PR to probe for cross platform native .dlls

Other changes are Jon's.


MLS.Agent/NativeAssemblyLoadHelper.cs
```diff
@@ -15,6 +15,7 @@ namespace MLS.Agent
 {
     public class NativeAssemblyLoadHelper : INativeAssemblyLoadHelper
     {
+        private static readonly HashSet<DirectoryInfo> globalProbingPaths = new HashSet<DirectoryInfo>();
```
KevinRansom: There are several instances of NativeAssemblyLoadHelper  created.  The globalProbingPaths  is the union of all of the located probing paths.

MLS.Agent/NativeAssemblyLoadHelper.cs
```diff
@@ -49,6 +54,119 @@ public void Handle(FileInfo assemblyFile)
             }
         }
 
+        private IEnumerable<string> ProbingFilenames(string name)
```
KevinRansom: the name passed in here is the value specified with the DllAttribute.  by convention, the dll may have:
1. A full path
2. A relativepath
3. lib prepended
4. on windows .dll or .exe as a suffix.
5. on linux .so as a suffix
6. on macos .dynlib as a suffix

This function enumerates the possibilities.

MLS.Agent/NativeAssemblyLoadHelper.cs
```diff
@@ -49,6 +54,119 @@ public void Handle(FileInfo assemblyFile)
...
+            bool ProbingUsePrefix(string name)
+            {
+                return name.IndexOf(Path.DirectorySeparatorChar) == -1              // If name has directory information no add no prefix
+                        && name.IndexOf(Path.AltDirectorySeparatorChar) == -1       // Alternate Directory seperator
+                        && name.IndexOf(Path.PathSeparator) == -1                   // Path seperator
+                        && name.IndexOf(Path.VolumeSeparatorChar) == -1;            // Volume seperator
+            }
+        }
+
+        private IEnumerable<string> ProbingPaths(string probingPath, string name)
```
KevinRansom: This function combines the probing root directory with the runtimes/rid/native to produce a list of dll's

MLS.Agent/NativeAssemblyLoadHelper.cs
```diff
@@ -64,28 +182,34 @@ private void OnAssemblyLoad(object sender, AssemblyLoadEventArgs args)
                 args.LoadedAssembly,
                 (libraryName, assembly, searchPath) =>
                 {
+                    var ptr = IntPtr.Zero;
                     if (_resolvers.TryGetValue(
```
KevinRansom: Use calling assembly path to limit probes to _probepaths.

MLS.Agent/NativeAssemblyLoadHelper.cs
```diff
@@ -1,50 +1,241 @@
...
+                        ptr = _probingPaths.SelectMany(di => ProbingPaths(di.FullName, libraryName).Select(dll => nativeLoader(dll))).FirstOrDefault();
+                    }
+                    if (ptr == IntPtr.Zero)
+                    {
+                       ptr = globalProbingPaths.SelectMany(di => ProbingPaths(di.FullName, libraryName).Select(dll => nativeLoader(dll))).FirstOrDefault();
+                    }
+                    return ptr;
+                });
+
+            IntPtr nativeLoader(string dll)
```
KevinRansom: If we can't load it, just return a Zero intptr

MLS.Agent/NativeAssemblyLoadHelper.cs
```diff
@@ -1,50 +1,241 @@
...
+                foreach(var rid in ProbingRids())
                 {
-                    NativeLibrary.SetDllImportResolver(args.LoadedAssembly, Resolve);
+                    var path = Path.Combine(probingPath, "runtimes", rid, "native");
+
+                    // Check it the dll exists in the rid specific native directory
+                    foreach (var pname in ProbingFilenames(name))
+                    {
+                        var p = Path.Combine(path, pname);
+                        if (File.Exists(p))
```
KevinRansom: Since we are probing for a file, don't return it if it isn't actually there.

NuGet.config
```diff
@@ -10,6 +10,6 @@
     <add key="dotnet-fsharp" value="https://dotnet.myget.org/F/fsharp/api/v3/index.json" />
     <add key="dotnet-try" value="https://dotnet.myget.org/F/dotnet-try/api/v3/index.json" />
     <add key="roslyn" value="https://dotnet.myget.org/F/roslyn/api/v3/index.json" />
-    <add key="dotnet-corefxlab" value="https://dotnet.myget.org/F/dotnet-corefxlab/api/v3/index.json" />
+    <add key="dotnet-corefxlab" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/MachineLearning/nuget/v3/index.json" />
```
jonsequitur: Did you mean to include this?

colombod: this file should not use azure package source

WorkspaceServer/Kernel/INativeAssemblyLoadHelper.cs
```diff
@@ -1,11 +1,17 @@
...
 namespace WorkspaceServer.Kernel
 {
-    public interface INativeAssemblyLoadHelper
+    public interface INativeAssemblyLoadHelper : IDisposable
     {
-        void Handle(string assembly);
-        void Configure(string v);
+        void Handle(ResolvedNugetPackageReference reference);
+
+        void SetNativeDllProbingPaths(IReadOnlyList<DirectoryInfo> nativeDllProbingPaths);
```
eerhardt: (nit) `NativeDll` may not be the best name here since `.dll` is a Windows concept/extension.

How about `NativeLibrary`?

MLS.Agent/NativeAssemblyLoadHelper.cs
```diff
@@ -1,50 +1,218 @@
...
+                            yield return p;
+                    }
+                }
+            }
+
+            // Computer valid dotnet-rids for this environment: https://docs.microsoft.com/en-us/dotnet/core/rid-catalog
+            // Where rid is: win, win-x64, win-x86, osx-x64, linux-x64 etc ...
+            IEnumerable<string> ProbingRids()
```
eerhardt: Why are we doing this ourselves? This is exactly what `AssemblyDependencyResolver` is meant to solve. It reads the `.deps.json` files and handles any RID fallbacks.

MLS.Agent/NativeAssemblyLoadHelper.cs
```diff
@@ -1,50 +1,218 @@
...
+            IntPtr nativeLoader(string dll)
             {
-                if (args.LoadedAssembly.Location == assembly)
+                var ptr = IntPtr.Zero;
+                try
                 {
-                    NativeLibrary.SetDllImportResolver(args.LoadedAssembly, Resolve);
+                    ptr = NativeLibrary.Load(dll);
+                    Logger.Log.Info("NativeLibrary.Load({dll})", args.LoadedAssembly.Location);
+                    Console.WriteLine($"NativeLibrary.Load({dll})");
```
eerhardt: Is this going to show up in the user's notebook output?

WorkspaceServer/Kernel/INativeAssemblyLoadHelper.cs
```diff
@@ -1,11 +1,17 @@
...
 
+using System;
+using System.Collections.Generic;
+using System.IO;
+using WorkspaceServer.Packaging;
+
 namespace WorkspaceServer.Kernel
 {
-    public interface INativeAssemblyLoadHelper
+    public interface INativeAssemblyLoadHelper : IDisposable
```
eerhardt: We made this `IDiposable`, but I don't see the instances being disposed of.

MLS.Agent/NativeAssemblyLoadHelper.cs
jonsequitur: We don't have a `.deps.json`. We doing a restore but not a build in order to avoid copying so many files out of the NuGet caches.

MLS.Agent/NativeAssemblyLoadHelper.cs
jonsequitur: Yes. The `Console.WriteLine` needs to be removed.

WorkspaceServer/Kernel/INativeAssemblyLoadHelper.cs
jonsequitur: It gets registered with the kernel:

https://github.com/dotnet/try/pull/600/files/775503234fee4b7966e947640b1fd1166b8ed347#diff-5b3346e53f8139b99a4f6237e72f4132R102

MLS.Agent/NativeAssemblyLoadHelper.cs
eerhardt: Maybe we should be generating a `.deps.json` file then. The logic being added here is super fragile and will not work in many cases.

For example, the proposed code won't handle the same scenario I just fixed in Azure Functions - https://github.com/Azure/azure-functions-host/issues/5187. In that case, the native library is located at `runtimes\{RID}\nativeassets\{TFM}`. You won't be able to guess at this path, unless you are planning on knowing how to fall back through TFMs. The current RID support here is also problematic because it won't allow for things like alpine to work.

The more we try to hard-code what the current system does, the more we are going to get it wrong. We should be using as much of the current systems as possible.

jonsequitur: > Maybe we should be generating a .deps.json file then.

Do you mean by doing a build or do you have another approach in mind?

eerhardt: You wouldn't need to do a full build to generate a `.deps.json` file. We have a few levels of places we can get at the current code.

1. Call into the [MSBuild Task directly](https://github.com/dotnet/sdk/blob/d8ef7ef571f035fa10f8e13935aa6afe54316579/src/Tasks/Microsoft.NET.Build.Tasks/GenerateDepsFile.cs#L19-L22).
2. Invoke the [MSBuild Target](https://github.com/dotnet/sdk/blob/be128d0ba81f0971ddbd5bb5a1039dd2c7768edf/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets#L158-L166)

Maybe @dsplaisted or @nguerrera have other ideas on how to generate a `.deps.json` file without doing a full build.

Another potential route would be to inspect NuGet's `project.assets.json` file and read all of the native asset paths from it. However, we would still need to know how to do RID fallback which the `AssemblyDependencyResolver` solves for us. But that is a solvable problem by reading the shared frameworks `.deps.json` file, which will give you the RID fallback graph for the current machine. (Assuming we are always running on a shared framework in dotnet-try.)

Also - another route is that if you just want to avoid "copying so many files out of the NuGet cache", you can set MSBuild properties to disable that. See https://github.com/dotnet/sdk/commit/1936ebd7307cebd6c5d1d8d699b9844f73b8fb2c.

> The `CopyLocalLockFileAssemblies` property can be set to `false` to revert to
the previous behavior for `dotnet build`.

dsplaisted: I agree with Eric that you should try to reuse as much of the existing build logic as possible.  I would avoid calling the `GenerateDepsFile` task directly, as a lot of the logic of what flows into that happens in targets files.  Running the `GenerateDepsFile` target is probably the right thing to do.

NotebookExamples/csharp/Samples/HousingML.ipynb
```diff
@@ -5,7 +5,7 @@
       "source": [
         "#r \"nuget:Microsoft.ML,1.4.0-preview\"\n",
         "#r \"nuget:Microsoft.ML.AutoML,0.16.0-preview\"\n",
-        "#r \"nuget:Microsoft.Data.DataFrame,0.1.1-e190920-1\""
+        "#r \"nuget:Microsoft.Data.DataFrame,1.0.0-e190910-1\""
```
eerhardt: Unfortunately we messed up some versions on our myget feed. The `1.0.0-` versions are older than the `0.1.1-` versions. If you are looking at the feed, note the date it was published.

The good news is that we have renamed the package to `Microsoft.Data.Analysis` and so all the versioning issues went away.

I can send a separate update to fix this to the latest package name and version, and fixing any code that needs updating.

jonsequitur: Please revert the notebook changes.

KevinRansom: @jonsequitur, I had already reverted them on my machine.  Eric, updated the version, and have updated it to the newer version.

The tests were failing because there was no way to specify the custom feed, with the new location for the project file.  I have added support for #r "nuget:restoresources=http://blah".

Unlike the F# repo, this needs to be specified before the #r nugets that rely on it.  So … first I suppose.

I am going to spend some time seeing if I can wire up the F# #r nuget processor, wich would enable me to rip out some of the code.

@eerhardt --- the primary benefit of reading the assets.json file, will be that there is less guesswork about where to find the file is, and so it should perform better in the resolution stage.  My main gripe, is there is no decent way to find the nuget root directory.


I will try to incorporate asset.json probing in the F# probing algorithm and thus the C# probing if I can successfully incorporate it.  However, using #I to specified probing roots for assemblies, has always been a thing in F# and we will want to continue with that including native probing, it's a nice developer scenario.

eerhardt: > My main gripe, is there is no decent way to find the nuget root directory.

I believe that is what the [`FallbackPackagePathResolver`](https://github.com/NuGet/NuGet.Client/blob/69a673582c62a63e917d152d871617ac5031dee1/src/NuGet.Core/NuGet.Packaging/FallbackPackagePathResolver.cs#L15) in NuGet is for. It will locate where a NuGet package is for a given context.

However, doesn't the Interactive Kernel control where the nuget root directory is?

jonsequitur: It doesn't.

jonsequitur: Per our discussions yesterday, I'd prefer to merge this PR and then revisit the approach, since there a number of different bugs being fixed here, not just the native dependency resolution.

PR #619, opened 2019-11-09
GettingStarted: Sessions: Fixed typo

talanc:
Should say "the `say_hello` region" instead of "the `hello` region"


talanc:
Will create another pull request

PR #671, opened 2019-11-27
split solutions

colombod:
split solutions and build


eng/build.ps1
```diff
@@ -12,7 +12,7 @@ function TestUsingNPM([string] $testPath) {
 
 try {
     # invoke regular build/test script
-    . (Join-Path $PSScriptRoot "common\build.ps1") @args
+    . (Join-Path $PSScriptRoot "common\build.ps1") "-projects $PSScriptRoot\..\dotnet-interactive.sln;$PSScriptRoot\..\dotnet-try.sln"
```
brettfo: You still have to pass through `@args`.

eng/build.sh
```diff
@@ -36,7 +36,7 @@ function TestUsingNPM() {
 }
 
 # invoke regular build/test script
-. "$scriptroot/common/build.sh" $args
+. "$scriptroot/common/build.sh" "$scriptroot/../dotnet-interactive.sln;$scriptroot/../dotnet-try.sln"
```
brettfo: Same, pass through `$args`.

eng/build.ps1
brettfo: Also, you _might_ have access to the variable `$RepoRoot` so you don't need `$PSScriptRoot\..`.

colombod: then why the line below in test like `TestUsingNPM "$PSScriptRoot\..\Microsoft.DotNet.Try.Client"`

PR #687, opened 2019-12-09
remove dependency on html agility pack in product code

colombod:
remove dependency on html agility pack in product code

### DIFF
--- a/XPlot.DotNet.Interactive.KernelExtensions/PlotlyChartExtensions.cs
+++ b/XPlot.DotNet.Interactive.KernelExtensions/PlotlyChartExtensions.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System.Text;
+using Microsoft.AspNetCore.Html;
 using XPlot.Plotly;
 using static Microsoft.DotNet.Interactive.Formatting.PocketViewTags;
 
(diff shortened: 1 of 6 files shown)


XPlot.DotNet.Interactive.KernelExtensions/PlotlyChartExtensions.cs
```diff
@@ -1,32 +1,28 @@
...
-            var divNode = document.DocumentNode.SelectSingleNode("//div");
-            var scriptNode = document.DocumentNode.SelectSingleNode("//script");
-
-            var newHtmlDocument = new HtmlDocument();
-
-            newHtmlDocument.DocumentNode.ChildNodes.Add(divNode);
-            newHtmlDocument.DocumentNode.ChildNodes.Add(GetScriptNodeWithRequire(scriptNode));
-
-            return newHtmlDocument.DocumentNode.WriteContentTo();
+            dynamic _ = new PocketView();
```
jonsequitur: If you `using static Microsoft.Dotnet.Interactive.Formatting.PocketViewTags` you can use `div`, etc. as top-level method calls and not have to declare a local `PocketView`.

XPlot.DotNet.Interactive.KernelExtensions/PlotlyChartExtensions.cs
```diff
@@ -1,32 +1,28 @@
...
+            var div = _.div[style: $"width: {chart.Width}px; height: {chart.Height}px;", id: chart.Id]();
+            var js = chart.GetInlineJS().Replace("<script>", string.Empty).Replace("</script>",string.Empty).Trim();
+            
+            return $@"{div}
```
jonsequitur: You don't need to use an interpolated string here, and in fact you're more likely to end up with HTML encoding issues if you do this.

This is the preferred approach:

```csharp
return div(GetScriptNodeWithRequire(js));
```

`GetScriptNodeWithRequire` should return `IHtmlContent` since it does not need to be encoded.

colombod: that will surrounds the div with another div. we were returning a different shape before

PR #693, opened 2019-12-12
Fix Issue #624 - #r nuget selects oldest package when version is not…

### LINKED ISSUE #624: #r loads the oldest NuGet from the feed
#### Describe the bug
Jupyter Lab / C# kernel, preview build of dotnet-try 1.0.19558.8 loads the newest (below v5.0.0) NuGet from the feed when the version is not defined in the #r command:
image

After upgrading to preview dotnet-try build 1.0.19558.9 the same #r command loads the oldest (below v1.0.0) NuGet package from the feed:
image

- [x] `dotnet-try`
- [ ] online

- [ ] User Interface (UI):  For example the output never displayed 
- [x] Other: The default NuGet package should be the newest one when version is not given in the #r command.

 - OS
    - [x] Windows 10
    - [ ] macOS
    - [ ] iOS
    - [ ] Android
 - Browser 
    - [x] Chrome
    - [ ] Edge
    - [ ] Safari

KevinRansom:
Fixes #624 - #r loads the oldest NuGet from the feed

By default when version is not specified nuget selects the oldest value in the feed.  This behaviour is not appropriate in a scripting environment when a developer is more likely to want the latest and greatest version.

This change uses "*"    select oldest stable version for the version value, when no version is specified.

### DIFF
--- a/Microsoft.DotNet.Interactive.Tests/LanguageKernelTests.cs
+++ b/Microsoft.DotNet.Interactive.Tests/LanguageKernelTests.cs
@@ -1556,7 +1556,7 @@ using Microsoft.ML.AutoML;
         }
 
         [Fact]
-        public async Task issue_637()
+        public async Task Pound_r_nuget_with_System_Text_Json_should_succeed()
         {
             var kernel = CreateKernel(Language.CSharp);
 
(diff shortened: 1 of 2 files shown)


Microsoft.DotNet.Interactive.Tests/LanguageKernelTests.cs
```diff
@@ -1576,5 +1576,27 @@ public async Task issue_637()
                   .ContainSingle(e => ((PackageAdded)e).PackageReference.PackageName == "Microsoft.NETCore.App.Ref");
 
         }
+
+        [Fact]
+        public async Task issue_624()
```
brettfo: More descriptive test name.  You can link to the fixed issue number in the attribute.

PR #696, opened 2019-12-12
Test case for System.Text.Json

### LINKED ISSUE #637: Unable to use System.Text.Json iin Jupyter c# kernel
#### Describe the bug
I just installed the c# jupyter kernel today for trying out. I can not get System.Text.Json namespace to import. If I do 
```using System.Text.Json;``` 
I get this error
 ```(1,19): error CS0234: The type or namespace name 'Json' does not exist in the namespace 'System.Text' (are you missing an assembly reference?)```
I also tried to reference it as a nuget package, with 
```#r "nuget:System.Text.Json"``` 
and it complete successful, but I still cannot use the namespace.

If I run ```Environment.Version``` from within the notebook I get 3.0.0 back. And ```dotnet --version``` shows 3.0.100 so I think I have the proper version installed.

I have installed some other packages, for example MongoDB.Driver could be installed with nuget and imported with using just fine, so it seems its something specifically with this package.

- [x] `dotnet-try`
- [ ] online

- [ ] User Interface (UI):  For example the output never displayed 
- [x] Other:

#### Screenshots
image

 - OS
    - [x] Windows 10
    - [ ] macOS
    - [ ] iOS
    - [ ] Android
 - Browser 
    - [ ] Chrome
    - [ ] Edge
    - [ ] Safari
    - [x] Firefox

### PR DESCRIPTION
**Pull request by KevinRansom**
Add a test case for using System.Text.Json.
Fixes #637 - Unable to use System.Text.Json iin Jupyter c# kernel #637

It now works, add a test case so that we keep it working.

### DIFF
--- a/Microsoft.DotNet.Interactive.Tests/LanguageKernelTests.cs
+++ b/Microsoft.DotNet.Interactive.Tests/LanguageKernelTests.cs
@@ -93,6 +93,44 @@ namespace Microsoft.DotNet.Interactive.Tests
                 .Be(5);
         }
 
+        // Option 1: inline switch
+        [Theory]
+        [InlineData(Language.CSharp)]
+        //[InlineData(Language.FSharp)]    Todo: Issue 695 - dotnet-interactive with an F# notebook does not load System.Text.Json #695
+        public async Task it_can_reference_system_text_json(Language language)
+        {
+            var source = language switch
+            {
+                Language.FSharp => new[]
+                {
+@"open System.Text.Json;
+let jsonException = JsonException()
+let location = jsonException.GetType().Assembly.Location
+location.EndsWith(""System.Text.Json.dll"")"
+                },
+
+                Language.CSharp => new[]
+                {
+@"using System.Text.Json;
+var jsonException = new JsonException();
+var location = jsonException.GetType().Assembly.Location;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: Microsoft.DotNet.Interactive.Tests/LanguageKernelTests.cs [hunk 1]
```diff
@@ -93,6 +93,44 @@ public async Task it_remembers_state_between_submissions(Language language)
...
+var location = jsonException.GetType().Assembly.Location;
+location.EndsWith(""System.Text.Json.dll"")"
+        }
+            };
+
+            var kernel = CreateKernel(language);
+
+            await SubmitCode(kernel, source);
+
+            KernelEvents
```
jonsequitur: You can also do this, which will be more informative if the event is missing:

```csharp
KernelEvents
  .Should()
  .ContainSingle<ReturnValueProduced>()
  .Which
  .Value
  .Should()
  .BeTrue();
```

KevinRansom: I like it ...

KevinRansom:
@brettfo, @jonsequitur,

sorry to be a pest can we merge this please, Github is not my friend.

Thanks

Kevin

PR #701, opened 2019-12-16
Add MyBinder link to example

dsyme:


dsyme:
If we add this, we should probably add it systematically to all examples both F# and C#

PR #715, opened 2019-12-29
Update NotebooksLocalExperience.md

TheDarkTrumpet:
Updated the kernel install instructions.  The current version didn't work, but the instructions on https://www.hanselman.com/blog/AnnouncingNETJupyterNotebooks.aspx do, so copied over some of those instructions here.


jonsequitur:
@TheDarkTrumpet Thanks! 

We didn't make this at all clear, but this change was intentional. We've moved development of these features over to `dotnet-interactive`. I updated the installation instructions to include the MyGet feed where the latest builds can be found.

PR #730, opened 2020-01-07
implement `%whos` magic command for F#

brettfo:
Most of this PR was simply moving the types `CurrentVariable` and `CurrentVariables` to a common location and also making common the type formatter.

image


Microsoft.DotNet.Interactive/CurrentVariables.cs
```diff
@@ -0,0 +1,72 @@
...
+        public CurrentVariables(IEnumerable<CurrentVariable> variables, bool detailed)
+            : this(detailed)
+        {
+            foreach (var variable in variables)
+            {
+                _variables[variable.Name] = variable;
+            }
+        }
+
+        public CurrentVariables(Dictionary<string, CurrentVariable> variables, bool detailed)
```
jonsequitur: This can be `IDictionary<string, CurrentVariable>`.

brettfo: Realized I don't use this overload anymore; I'll just get rid of it.

Microsoft.DotNet.Interactive/CurrentVariables.cs
```diff
@@ -0,0 +1,72 @@
...
+            Detailed = detailed;
+        }
+
+        public bool Detailed { get; }
+
+        public IEnumerator<CurrentVariable> GetEnumerator() => _variables.Values.GetEnumerator();
+
+        IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
+
+        public static void RegisterFormatter()
```
jonsequitur: Rather than having this work via a side effect, it might be better to have a method that returns an instance of `ITypeFormatter<CurrentVariables>`.

brettfo: Added a separate `ITypeFormatter<CurrentValue>`.  Let me know if that's what you had in mind.

Microsoft.DotNet.Interactive/CurrentVariables.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    public class CurrentVariables : IEnumerable<CurrentVariable>
+    {
+        private readonly Dictionary<string, CurrentVariable> _variables = new Dictionary<string, CurrentVariable>();
+
+        public CurrentVariables(IEnumerable<CurrentVariable> variables, bool detailed)
+            : this(detailed)
+        {
+            foreach (var variable in variables)
+            {
```
jonsequitur: This could use a null check.

Microsoft.DotNet.Interactive/CurrentVariables.cs
jonsequitur: That works. There's also `Formatter.Create` if you prefer to avoid the ceremony of creating a class.

Microsoft.DotNet.Interactive/CurrentVariables.cs
jonsequitur: Even better.

PR #731, opened 2020-01-07
Added dispose pattern and overrides Finalize on KernelBase

TIHan:
This just adds the dispose pattern and overrides `Finalize` on `KernelBase`. Disposing on `Finalize` is basically a fallback for the kernel.

### DIFF
--- a/Microsoft.DotNet.Interactive/KernelBase.cs
+++ b/Microsoft.DotNet.Interactive/KernelBase.cs
@@ -46,6 +46,11 @@ namespace Microsoft.DotNet.Interactive
             _disposables.Add(_kernelEvents);
         }
 
+        ~KernelBase()
+        {
+            Dispose(false);
+        }
+
         public KernelCommandPipeline Pipeline { get; }
 
         private void AddSetKernelMiddleware()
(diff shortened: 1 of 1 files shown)


Microsoft.DotNet.Interactive/KernelBase.cs
```diff
@@ -251,7 +256,21 @@ public void RegisterForDisposal(IDisposable disposable)
             IKernelCommand command,
             KernelInvocationContext context) => context.HandlingKernel = this;
 
-        public void Dispose() => _disposables.Dispose();
+        protected virtual void Dispose(bool disposing)
+        {
+            if (_disposables.IsDisposed)
```
eerhardt: Should `_disposables` be responsible for checking `IsDisposed` in its `Dispose()` method, and no-oping? That way `KernelBase` can just call `Dispose()` on the things it needs to dispose.

Microsoft.DotNet.Interactive/KernelBase.cs
```diff
@@ -251,7 +256,21 @@ public void RegisterForDisposal(IDisposable disposable)
...
+            {
+                return;
+            }
+
+            _disposables.Dispose();
```
eerhardt: This should only be called when `disposing` is `true`. When you are finalizing `KernelBase`, you shouldn't be touching other managed objects. 

See https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose?view=netframework-4.8#the-disposeboolean-overload

Microsoft.DotNet.Interactive/KernelBase.cs
TIHan: It probably does, I'm doing this as a sanity check.

Microsoft.DotNet.Interactive/KernelBase.cs
TIHan: Yes that is right, but we are disposing unmanaged bits in this since we register disposals on inherited classes such as FSharpKernel which will register to dispose the generated assemblies.

TIHan: FSharpKernel arguably should be following the correct dispose pattern, which I know for a fact we don't :)

TIHan:
Closing as I think my use case of making sure our Fsi session resources get freed really needs to not be handled by Finalize.

PR #734, opened 2020-01-08
move to .NET SDK 3.1.100

### LINKED ISSUE #733: Errors when using ML.NET on a machine with only .NET Core 3.1 installed
#### Describe the bug
On a machine with only .NET Core 3.1 installed (note it can't have .NET Core 3.0 and 3.1) - run the following notebook:

```C#
#r "nuget:Microsoft.ML,1.4.0"
```

```C#
using Microsoft.ML;
var mlContext = new MLContext(seed: 1);
```

I expect this to work correctly, however it fails with strange errors:

System.TypeLoadException: Could not load type 'System.Nullable`1' from assembly 'ℛ*7375a6bc-9346-4650-9f1b-255a041c874c#1-25,

- [ ] `dotnet-try`
- [x] .NET Jupyter Notebook 
- [ ] online

- [x] User Interface (UI):  For example the output never displayed 
- [ ] Other:

#### Screenshots
image

 - OS
    - [x] Windows 10
    - [ ] macOS
    - [ ] iOS
    - [ ] Android
 - Browser 
    - [x] Chrome
    - [ ] Edge
    - [ ] Safari

jonsequitur:
This fixes #733.

### DIFF
--- a/eng/common/templates/job/performance.yml
+++ b/eng/common/templates/job/performance.yml
@@ -7,7 +7,7 @@ parameters:
   container: ''                   # required -- name of the container
   osGroup: ''                     # required -- operating system for the job
   extraSetupParameters: ''        # optional -- extra arguments to pass to the setup script
-  frameworks: ['netcoreapp3.0']   # optional -- list of frameworks to run against
+  frameworks: ['netcoreapp3.1']   # optional -- list of frameworks to run against
   continueOnError: 'false'        # optional -- determines whether to continue the build if the step errors
   dependsOn: ''                   # optional -- dependencies of the job
   timeoutInMinutes: 320           # optional -- timeout for the job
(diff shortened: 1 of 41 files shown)


eng/common/templates/job/performance.yml
```diff
@@ -7,7 +7,7 @@ parameters:
   container: ''                   # required -- name of the container
   osGroup: ''                     # required -- operating system for the job
   extraSetupParameters: ''        # optional -- extra arguments to pass to the setup script
-  frameworks: ['netcoreapp3.0']   # optional -- list of frameworks to run against
+  frameworks: ['netcoreapp3.1']   # optional -- list of frameworks to run against
```
brettfo: Everything under `eng/common` is owned by Arcade.  Soon we'll get an auto-PR that reverts this line.  No idea on what'll happen, though.

jonsequitur: Oops, didn't notice this. Any concern that it'll cause misbehavior?

PR #739, opened 2020-01-10
share ncrunch configuration

colombod:


dotnet-interactive.v3.ncrunchsolution
```diff
@@ -0,0 +1,6 @@
+﻿<SolutionConfiguration>
+  <Settings>
+    <AllowParallelTestExecution>False</AllowParallelTestExecution>
```
jonsequitur: Have you run into any issues with parallel execution that we're not handling directly in the affected tests? Typically it's better to leave this on true.

colombod: Not at all, I disabled this when trying to figure out why this was broken. Will enable and push

PR #740, opened 2020-01-12
Make `build-and-install-dotnet-interactive.ps1` work on Linux and macOS with pwsh

daxian-dbw:
Make `build-and-install-dotnet-interactive.ps1` work on Linux and macOS with `pwsh` (the cross-plat PowerShell).

`$IsLinux` and `$IsMacOS` are the built-in readonly variables defined in `pwsh`.


jonsequitur:
Thanks @daxian-dbw!

Issue #729, opened 2020-01-07
Extend dotnet try global tool to copy code blocks from sample solution to markdown files


ax0l0tl:
**Problem**
I want to use .net interactive documentation to write my documentation but markdown doc files should still display my code samples when viewed in an environment without try .net hosting (git hub README or simple markdown viewer). Right now I have to copy the sample regions manually  from the C# projects to markdcown code blocks to make sure the samples are visible without try .net, which is error-prone and inconvenient.

**Suggestion**
I would like to use the global dotnet try tool to keep my markdown in sync with the sample code. It would be great to have a *sync* command that just copies the referenced regions into the markdown code blocks. Usage would be similar to the existing verify command:

dotnet try sync <RootDirectory>

Even if the docs are never used as interactive docs one gets the great advantage of writing sample code with compiler support. So no more broken code samples in the documentation.

**Basic implementation**
I created a basic implementation with the current try dotnet version. It is working for me and perhaps it is useful to start with (unfortunately pull requests are not possible at the moment):

``` csharp

using System.CommandLine;
using System.IO;
using System.Linq;
using System.Threading.Tasks;
using Markdig;
using Markdig.Renderers.Normalize;
using Markdig.Syntax;
using Microsoft.DotNet.Try.Markdown;
using MLS.Agent.Markdown;
using MLS.Agent.Tools;
using WorkspaceServer;

namespace MLS.Agent.CommandLine
{
    public static class SyncCommand
    {
        public static async Task<int> Do(
            SyncOptions syncOptions,
            IConsole console,
            StartupOptions startupOptions = null
        )
        {
            var directoryAccessor = syncOptions.RootDirectory;
            var packageRegistry = PackageRegistry.CreateForTryMode(directoryAccessor);
            var markdownProject = new MarkdownProject(
                directoryAccessor,
                packageRegistry,
                startupOptions);

            var markdownFiles = markdownProject.GetAllMarkdownFiles().ToArray();
            if (markdownFiles.Length == 0)
            {
                console.Error.WriteLine($"No markdown files found under {directoryAccessor.GetFullyQualifiedRoot()}");
                return -1;
            }

            foreach (var markdownFile in markdownFiles)
            {
                var document = ParseMarkdownDocument(markdownFile);

                var pipeline = new MarkdownPipelineBuilder().UseNormalizeCodeBlockAnnotations().Build();
                var writer = new StringWriter();
                var renderer = new NormalizeRenderer(writer);
[...]

Issue #214, opened 2019-05-18
support for publishing standalone Markdown, static HTML, WebAssembly hosts


jonsequitur:
Some aspects of the Try .NET Markdown format make the source Markdown file less than ideal for display directly within e.g. a GitHub `readme.md`:

* Empty code fences referencing a backing project. These are populated at runtime by `dotnet try` but other Markdown parsers will render them as empty. While the content writer can populate them by copy-pasting from the source project, we want to avoid this maintenance burden.
* #187: file-wide settings.

Additionally, we should support static site generators and content management systems by doing the Markdown-to-HTML conversion at publish time rather than expecting it to be done at runtime.

The following command line examples are just sketches. In all cases, new files are emitted to `<output dir>` and source Markdown files are left unchanged.

- [ ] `dotnet try publish --format Markdown --interactive:false <output dir>`: 
    * Emits a Markdown file for each source Markdown file. 
    * Populates code fences using the latest backing source in a format suitable for use as read-only, non-interactive documentation, e.g. in GitHub's readme or wiki.
- [ ] `dotnet try publish --format Markdown --interactive:true <output dir>`: 
    * Emits a Markdown file for each source Markdown file.
    * Emits a Try .NET WebAssembly package, which contains a Blazor app capturing the backing project's NuGet dependencies. 
    * Populates code fences using the latest backing source. 
    * Replaces `--project` annotations with `--package` annotations referencing to the Try .NET WebAssembly package.
- [ ] `dotnet try publish --format HTML --interactive:false <output dir>` 
    * Emits an HTML file for each source Markdown file.
    * Transforms Markdown, replacing code fences with Try .NET-attributed `<pre><code>` blocks containing the latest backing source.
- [ ] `dotnet try publish --format HTML --interactive:true <output dir>` 
    * Emits an HTML file for each source Markdown file.
    * Creates a Try .NET WebAssembly package. 
    * Transforms Markdown into HTML, replacing code fences with Try .NET-attributed `<pre><code>` blocks containing the latest backing source. 
    * Replaces `data-trydotnet-package` attribute values with refences to the Try .NET WebAssembly package.

akshita31:
This looks really good. A couple questions:
What will be a general usage pattern for the publish step. 
1. Like is it my responsibility as a content author that whenever I make some changes to my markdown file or to my backing project, I should run the command and then submit the PR ?
2. Or, it will be a task in the build pipeline that will execute everytime a PR is submitted ?

[...]

PR #741, opened 2020-01-13
Dotnet try publish

ax0l0tl:
This is a proposal for #729 (#214).

Usage would be:
`publish "..\Samples" --format Markdown --target-directory "..\Samples\published"`

Supported formats are Markdown and HTML

Root directory can be identical to target-directory. In that case with Markdown formatting selected, source files will replaced with the ones that contain the updated source blocks from the sample solutions.

I was insecure about file access. Read access to directories is done via IDirectoryAccessor but write access is not abstracted, that seems incorrect. So any suggestions welcome.


ax0l0tl:
I don't know why the try-ci Kernel_can_pound_r_nuget_using_kernel_client test is failing. As far as I know I did not touch anything in this area and it actually runs on my machine. Could anyone help?

jonsequitur:
Thanks for working on this, @ax0l0tl!

The test is just being flaky. We're in the process of splitting up the dotnet-try and dotnet-interactive solutions into separate repositories. The solutions are already separated other than a few of the build infrastructure pieces, so if you see tests failing for dotnet-interactive, it's almost definitely not because of a change you made.

MLS.Agent.Tools/RelativePath.cs
```diff
@@ -189,4 +189,20 @@ protected static void ThrowIfContainsDisallowedFilePathChars(string filename)
         //     return !Equals(left, right);
         // }
     }
+
+    public static class RelativePathExtension
```
jonsequitur: We typically put classes into their own files.

Microsoft.DotNet.Try.Markdown/NormalizeBlockAnnotationExtension.cs
```diff
@@ -0,0 +1,83 @@
+using Markdig;
+using Markdig.Renderers;
+using Markdig.Renderers.Normalize;
+using Markdig.Syntax;
+
+namespace Microsoft.DotNet.Try.Markdown
```
jonsequitur: Please split these into separate files per class.

MLS.Agent.Tools/RelativeDirectoryPath.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace MLS.Agent.Tools
 {
-    public class RelativeDirectoryPath :
+    public sealed class RelativeDirectoryPath :
```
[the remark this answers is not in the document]
ax0l0tl: Your're right, no need to enforce seal here. Just wanted to assert that no unknown direct subclass of RelativePath is created (because we switch on RelativeFilePath and RelativeDirectoryPath), but this is already asserted by the internal constructor of RelativePath.

MLS.Agent.Tests/CommandLine/PublishCommandTests.cs
```diff
@@ -0,0 +1,70 @@
...
+
+                var directoryAccessor = new InMemoryDirectoryAccessor(rootDirectory)
+                {
+                    ("doc.md", markdown)
+                }.CreateFiles();
+
+
+                var targetDirectory = await DoPublish(directoryAccessor);
+
+                var files = targetDirectory.GetAllFilesRecursively().ToList();
```
jonsequitur: Do you want to assert something about the files?

MLS.Agent.Tests/CommandLine/PublishCommandTests.cs
```diff
@@ -0,0 +1,70 @@
...
+            }
+
+            private async Task<InMemoryDirectoryAccessor> DoPublish(IDirectoryAccessor directoryAccessor, IDirectoryAccessor target = null)
```
jonsequitur: It would help to clarify the purpose of the two `IDirectoryAccessor` parameters. Would `source` and `target` would make sense?

And did you mean to use `target` rather than initializing a new `targetDirectory` below?

MLS.Agent.Tools/RelativePath.cs
```diff
@@ -11,7 +11,7 @@ public abstract class RelativePath
     {
         private string _value;
 
-        protected RelativePath(string value)
+        protected internal RelativePath(string value)
```
jonsequitur: This change should be unnecessary.

MLS.Agent.Tests/CommandLine/PublishCommandTests.cs
```diff
@@ -36,20 +49,69 @@ public class WithMarkdownOutputFormat
...
+")]
+            public async Task When_there_are_code_fence_annotations_in_markdown_content_of_the_fenced_section_is_replaced_with_the_project_code(string markdown)
+            {
+                var files = PrepareFiles(
+                    ("./folder/project/some.csproj", CsprojContents),
+                    ("./folder/project/Program.cs", CompilingProgramWithRegionCs),
+                    ("./folder/doc.md", markdown)
+                );
+
+                var (publishOutput, resultCode) = await DoPublish(files).ConfigureAwait(false);
```
jonsequitur, line 82: You shouldn't need `ConfigureAwait` in the tests. We're not doing this elsewhere. If we're seeing a deadlock, there's a bug somewhere in our product code and we can track it down.

jonsequitur:
@ax0l0tl This is looking good. I have a couple of comments on tests. Do you feel like it's ready to merge?

MLS.Agent.Tools/RelativePath.cs
ax0l0tl: Removed internal. The idea was to avoid unknown subclasses of RelativePath, because we rely on knowing the subtypes, but since this is not an API assembly, I aggree it is unnecessary.

ax0l0tl:
@jonsequitur I added a test to assert file handling is correct in case of target is subdirectory of source. From my side merging would be fine!

MLS.Agent/CommandLine/CommandLineParser.cs
```diff
@@ -435,6 +448,27 @@ Command Verify()
...
+            {
+                var publishCommand = new Command("publish", "Publish code from sample projects to Markdown files in the target directory and its children.")
+                {
+                    new Option("--format", "Format of the files to publish")
+                    {
+                        Argument = new Argument<PublishFormat>()
+                    },
+                    new Option("--target-directory", "Specify the path where the rendered files should go. Can be equal to root directory. In case of Markdown output format and equal root and target directories, original source files will be replaced.")
+                    {
+                        Argument = new Argument<FileSystemDirectoryAccessor>(() => new FileSystemDirectoryAccessor(Directory.GetCurrentDirectory()))
```
jonsequitur, line 462: I suspect this argument should be a `DirectoryInfo`.

PR #747, opened 2020-01-15
Extension loading Refactoring

colombod:


Microsoft.DotNet.Interactive.Tests/CompositeKernelTests.cs
```diff
@@ -70,6 +71,95 @@ public async Task Handling_kernel_can_be_specified_using_kernel_name_as_a_magic_
...
+            }
+            
+            using var events = kernel.KernelEvents.ToSubscribedList();
+
+            await kernel.SendAsync(
+                new SubmitCode(
+                    @"var x = 123;",
+                    "unregistered kernel name"));
+
+            events.Should().ContainSingle<CommandFailed>();
```
jonsequitur: We should verify the nature of the failure as well to avoid a false positive.

PR #753, opened 2020-01-16
track events without middleware

colombod:


Microsoft.DotNet.Interactive/CompositeKernel.cs
```diff
@@ -24,27 +24,20 @@ public CompositeKernel()
...
-                {
-                    await next(command, context);
-                }
-            });
+        private void InterceptAddPackageCommand()
+        {
+            RegisterForDisposal(KernelEvents.OfType<PackageAdded>()
+                .Select(pa => pa.PackageReference.PackageRoot)
+                .Where(root => root?.Exists == true)
+                .Distinct()
```
jonsequitur: Nice way to dedupe for the life of the kernel!

PR #757, opened 2020-01-18
update System.CommandLine.Experimental -> System.CommandLine

jonsequitur:


Microsoft.DotNet.Interactive.CSharp/CSharpKernelExtensions.cs
```diff
@@ -44,8 +45,20 @@ public static class CSharpKernelExtensions
 
         public static CSharpKernel UseNugetDirective(this CSharpKernel kernel)
         {
-            var packageRefArg = new Argument<PackageReference>((SymbolResult result, out PackageReference reference) =>
-                                                                        PackageReference.TryParse(result.Token.Value, out reference))
+            var packageRefArg = new Argument<PackageReference>(
```
KathleenDollard: This is totally a stylistic statement that can be ignored, but I want to share it. 

It's _quite_ grating to me when the logic of an operation appears in the declaration. When i want to see a CLI declaration, I do not care about what is happening except in intent (by method name). Are we using C#8? Can we use local functions for this?

jonsequitur: For this many lines of code, agreed.

PR #759, opened 2020-01-21
memory optimisation

colombod:


dotnet-interactive.Profiler/Program.cs
```diff
@@ -0,0 +1,70 @@
...
+using System.Threading.Tasks;
+using JetBrains.Profiler.Api;
+using Microsoft.DotNet.Interactive;
+using Microsoft.DotNet.Interactive.App;
+using Microsoft.DotNet.Interactive.Commands;
+using Microsoft.DotNet.Interactive.CSharp;
+using Microsoft.DotNet.Interactive.FSharp;
+using Microsoft.DotNet.Interactive.Jupyter;
+
+namespace dotnet_interactive.Profiler
```
jonsequitur: The root namespace should be `Microsoft.DotNet.Interactive.App`.

colombod: done

Microsoft.DotNet.Interactive.CSharp/CSharpKernel.cs
```diff
@@ -66,6 +65,10 @@ public CSharpKernel()
             _extensionLoader = new CSharpKernelExtensionLoader();
             NativeAssemblyLoadHelper = new NativeAssemblyLoadHelper(this);
             RegisterForDisposal(NativeAssemblyLoadHelper);
+            RegisterForDisposal(() =>
+            {
+                ScriptState = null;
```
colombod: @tmat is there any better and more deterministic way to dispose of ScriptState?

PR #760, opened 2020-01-21
renamed tool and docker files

colombod:


brettfo:
As noted in the build logs, you need to update the file patterns in `eng/AfterSolutionBuild.targets`.

PR #761, opened 2020-01-22
handling kernel captured before code splitter list execution

colombod:


Microsoft.DotNet.Interactive/KernelBase.cs
```diff
@@ -105,7 +105,7 @@ command switch
                     }
                     else
                     {
-                        await context.HandlingKernel.SendAsync(command);
+                        await (handlingKernel ?? context.HandlingKernel).SendAsync(command);
```
brettfo: This concerns me, I'd be more interested to see `if (handlingKernel != null && handlingKernel != context.HandlingKernel)` and then debug back to see why/how it changed.

colombod: A good point and will remove that null check as I think that would be a bug

colombod: Actually that is needed as handling kernel can be set via directives and that is a case where the handling kernel at the beginning of the method is null

PR #858, opened 2020-07-13
always run restore when building with arcade

brettfo:
These scripts are only run during a local build, but it should make it easier to fall into the pit of success.


07691 (commented):
Think  forward. Codes

PR #861, opened 2020-07-16
Fill console fences on publish

jonsequitur:
dotnet-try-publish


jonsequitur:
@LadyNaggaga  @BillWagner

PR #871, opened 2020-08-04
[WIP ] update blazor

colombod:


MLS.Agent/Properties/launchSettings.json
```diff
@@ -36,7 +36,7 @@
     },
     "Verify": {
       "commandName": "Project",
-      "commandLineArgs": "verify \"..\\Samples\""
+      "commandLineArgs": "verify \"C:\\Users\\dicolomb\\source\\repos\\csharp8\""
```
jonsequitur, line 39: We should probably not even check in this file.

colombod: reverted

PR #963, opened 2021-09-29
Switch to 1ES R&D pools on main

jonfortescue:
We need to switch to the new 1ES hosted pools. This does that.

cc/ @ulisesh @lpatalas


jonsequitur (approved):

jonfortescue:
@jonsequitur Could you merge this? I don't have permission to do so. :)

PR #1056, opened 2022-12-14
Fix demands in azure-pipelines.yml

akoeplinger:
It needs the ImageOverride otherwise it'll fall back to some default AzDO image


akoeplinger:
@colombod ping

colombod:
@brettfo need your help!!!

brettfo (approved):

PR #1112, opened 2024-02-08
[main] Fix - Update dependencies from dotnet/arcade

ocallesp:
This is a fix for https://github.com/dotnet/try/pull/1108


ocallesp:
In the internal build it failed with the following error message:

```
Build FAILED.

D:\a\_work\_tool\dotnet\sdk\8.0.101\Microsoft.Common.CurrentVersion.targets(4911,5): error MSB3030: Could not copy the file "D:\a\_work\1\s\artifacts\bin\Microsoft.TryDotNet\Release\net8.0\Microsoft.TryDotNet.pdb" because it was not found. [D:\a\_work\1\s\src\Microsoft.TryDotNet.IntegrationTests\Microsoft.TryDotNet.IntegrationTests.csproj]
D:\a\_work\_tool\dotnet\sdk\8.0.101\Microsoft.Common.CurrentVersion.targets(4911,5): error MSB3030: Could not copy the file "D:\a\_work\1\s\artifacts\bin\Microsoft.TryDotNet.WasmRunner\Release\net8.0\Microsoft.TryDotNet.WasmRunner.pdb" because it was not found. [D:\a\_work\1\s\src\Microsoft.TryDotNet.IntegrationTests\Microsoft.TryDotNet.IntegrationTests.csproj]
    0 Warning(s)
    2 Error(s)
```

AbhitejJohn:
@ocallesp : do you know if we are building with embedded pdbs? That is the default AFAIK, in which case, I wouldn't expect those files to be present. We'd want to fix our code.

On a side note, did you want to publish this PR?

ocallesp:
> @ocallesp : do you know if we are building with embedded pdbs? That is the default AFAIK, in which case, I wouldn't expect those files to be present. We'd want to fix our code.
> 
> On a side note, did you want to publish this PR?

None of the projects are using embedded PDBs, so we are expecting the .pdb files to be generated.

ocallesp:
Passed internally
https://dev.azure.com/dnceng/internal/_build/results?buildId=2386862&view=results

src/Microsoft.TryDotNet.IntegrationTests/EditorTests.cs
```diff
@@ -356,6 +357,7 @@ public async Task when_user_code_in_editor_diagnostics_are_produced()
     }
 
     [IntegrationTestFact]
+    [Obsolete]
```
AbhitejJohn, line 360: Curious, why did we have to make these obsolete? Are these tests failing?

jonsequitur: If this is to prevent the warning that the tests are using an obsolete API, the preferable approach is to suppress that warning for the entire test project, since it's not really adding value here.

jonsequitur (approved):

PR #1113, opened 2024-02-08
Publish to public/dotnet-tools when internal azure

ocallesp:


AbhitejJohn:
This isn't required anymore given our new deployment strategy.

Issue #1119, opened 2024-02-29
update .net interactive packages and npm


colombod:
Update to latest .NET Interactive and npm 
Check for npm dependencies update and the monaco component integration versions

PR #1125, opened 2024-03-05
Update CSharpProject and npm packages

ocallesp:
I will submit updates in parts so we don't break the whole code.

This is part of https://github.com/dotnet/try/issues/1119

### DIFF
--- a/src/microsoft-trydotnet-editor/package.json
+++ b/src/microsoft-trydotnet-editor/package.json
@@ -56,7 +56,8 @@
   "dependencies": {
     "@microsoft/polyglot-notebooks": "^1.0.515601",
     "monaco-editor": "0.33.0",
-    "rxjs": "7.5.5"
+    "rxjs": "7.5.5",
+    "uuid": "7.0.0"
   },
   "mocha": {
     "ui": "bdd",
--- a/src/microsoft-trydotnet-editor/package.json
+++ b/src/microsoft-trydotnet-editor/package.json
@@ -50,7 +50,7 @@
     "ts-mocha": "10.0.0",
     "ts-node": "10.8.1",
     "typescript": "4.7.4",
-    "webpack": "5.73.0",
+    "webpack": "^5.90.3",
     "webpack-cli": "4.10.0"
   },
   "dependencies": {
--- a/src/microsoft-trydotnet-editor/package.json
+++ b/src/microsoft-trydotnet-editor/package.json
@@ -54,7 +54,7 @@
     "webpack-cli": "4.10.0"
   },
   "dependencies": {
-    "@microsoft/polyglot-notebooks": "1.0.441801",
+    "@microsoft/polyglot-notebooks": "^1.0.515601",
     "monaco-editor": "0.33.0",
     "rxjs": "7.5.5"
   },
--- a/src/microsoft-trydotnet-editor/package.json
+++ b/src/microsoft-trydotnet-editor/package.json
@@ -54,7 +54,7 @@
     "webpack-cli": "4.10.0"
   },
   "dependencies": {
-    "@microsoft/polyglot-notebooks": "^1.0.515601",
+    "@microsoft/polyglot-notebooks": "1.0.441801",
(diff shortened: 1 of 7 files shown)


colombod:
we need to update also the npm package here https://github.com/dotnet/try/blob/bb94a0819d6a23c7797e235b6c1b58da438eb55b/src/microsoft-trydotnet-editor/package.json#L57

ocallesp:
> we need to update also the npm package here
> 
> https://github.com/dotnet/try/blob/bb94a0819d6a23c7797e235b6c1b58da438eb55b/src/microsoft-trydotnet-editor/package.json#L57

polyglot-notebooks updated

AbhitejJohn:
@ocallesp: I'm curious, any reason why we aren't doing an upgrade of all the packages that need a change? Reviewing multiple PRs for each package upgrade sounds like a lot of work. And that's likely what @colombod is suggesting above.

ocallesp:
> @ocallesp: I'm curious, any reason why we aren't doing an upgrade of all the packages that need a change? Reviewing multiple PRs for each package upgrade sounds like a lot of work. And that's likely what @colombod is suggesting above.

Doing it in parts. I upgraded all of them and broke the build.

colombod:
> @ocallesp: I'm curious, any reason why we aren't doing an upgrade of all the packages that need a change? Reviewing multiple PRs for each package upgrade sounds like a lot of work. And that's likely what @colombod is suggesting above.

He si doing the right job, but some packages are linked, upgrading .NET Interactive Nuget without upgrading the npm too is a problem as they are linked by the protocol objects.

src/microsoft-trydotnet-editor/package.json
```diff
@@ -50,13 +50,14 @@
     "ts-mocha": "10.0.0",
     "ts-node": "10.8.1",
     "typescript": "4.7.4",
-    "webpack": "5.73.0",
+    "webpack": "^5.90.3",
```
AbhitejJohn, line 53: From what I know, we have been using fixed versions for compliance reasons. We'd likely want to preserve that here as well.

src/microsoft-trydotnet-editor/package.json
```diff
@@ -50,13 +50,14 @@
...
     "webpack-cli": "4.10.0"
   },
   "dependencies": {
-    "@microsoft/polyglot-notebooks": "1.0.441801",
+    "@microsoft/polyglot-notebooks": "^1.0.515601",
```
AbhitejJohn, line 57: Similar comment about versioning here as well.

src/microsoft-trydotnet-editor/tests/apiServiceSimulator.ts
```diff
@@ -48,7 +48,6 @@ function patchEvents(events: polyglotNotebooks.KernelEventEnvelopeModel[], comma
         let event = events[i];
         var sourceCommand = commands.find(command => command.commandType === event.command!.commandType);
         let patchedEvent = { ...event };
-        patchedEvent.command!.id = sourceCommand!.id;
```
AbhitejJohn, line 51: Could you share some context on why this was removed please?

AbhitejJohn:
Tagging @jonsequitur / @colombod for approval.

src/microsoft-trydotnet-editor/tests/apiServiceSimulator.ts
ocallesp: this concept was removed.

src/microsoft-trydotnet-editor/package.json
ocallesp: Cannot revert this change to 5.73.0 because is a security vulnerability in webpack in the range of 5.0.0 - 5.75.0

src/microsoft-trydotnet-editor/package.json
ocallesp: reverted this change

colombod (approved):

PR #1149, opened 2024-03-29
Update Peaky tests, prebuild console package on app startup

jonsequitur:

PR #1161, opened 2024-04-09
More attempts to get more test logging

AbhitejJohn:
This is based on changes in this PR: https://github.com/dotnet/try/pull/1149


azure-pipelines.yml
```diff
@@ -131,6 +133,7 @@ stages:
           condition: ne(variables['SkipTests'], 'true')
           env:
             RunIntegrationTests: $(RunIntegrationTests)
+            POCKETLOGGER_LOG_PATH: $(PocketLoggerLogPath)
```
jonsequitur, line 136: We need this set as well for the unit tests as well as for the child process (the `AspNetProcess` class) spawned by the integration tests.

Setting it globally is really the simplest approach.

AbhitejJohn: Cool, my latest push should handle setting that more globally.

Issue #1150, opened 2024-04-04
Consolidate fixes in our yml pipelines


AbhitejJohn:
Currently any fixes in the yml needs updates in 3 different places:
1. https://github.com/dotnet/try/blob/main/azure-pipelines-CI.yml
2. https://github.com/dotnet/try/blob/main/azure-pipelines.yml
3. https://devdiv.visualstudio.com/DefaultCollection/Engineering/_apps/hub/ms.vss-build-web.ci-designer-hub?pipelineId=21351&branch=InitialSetup

That's a good amount of yml duplication. Investigate if we can use a template that we can pull into all these yml files to make this easier.

ocallesp:
Doing some refactor only:
- https://github.com/dotnet/try/blob/main/azure-pipelines-CI.yml
- https://github.com/dotnet/try/blob/main/azure-pipelines.yml

PR #1162, opened 2024-04-18
Refactor pipelines to remove duplicate code

ocallesp:
We are only making changes to these two pipelines:
- https://github.com/dotnet/try/blob/main/azure-pipelines-CI.yml
- https://github.com/dotnet/try/blob/main/azure-pipelines.yml

This part of [Consolidate fixes in our yml pipeline](https://github.com/dotnet/try/issues/1150)

### DIFF
--- a/eng/templates/setup-template.yml
+++ b/eng/templates/setup-template.yml
@@ -1,0 +1,19 @@
+# setup-template.yml
+steps:
+- script: git config --global core.longpaths true
+  displayName: Enable `git clean` to handle long paths
+
+- checkout: self
+  clean: true
+
+- task: NodeTool@0
+  displayName: Add NodeJS/npm
+  inputs:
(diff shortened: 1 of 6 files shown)


eng/templates/setup-template.yml
```diff
@@ -0,0 +1,19 @@
+# setup-template.yml
```
AbhitejJohn, line 1: Could we add a comment to say that this affects both windows and linux builds here please? It should be apparent from code but wouldn't hurt to have multiple indicators.

AbhitejJohn:
Was this primarily re-arranging existing tasks or did we have to add/remove tasks as well?

AbhitejJohn (approved):
LGTM, will let Dale/Jon who have been spending time here take a stab at a review as well though.

dalehhirt:
Talked with Osvaldo, he is working on updating this

PR #1168, opened 2024-04-19
update dev guide and peaky test notebook

jonsequitur:


AbhitejJohn:
@ocallesp : When this is ready, any chance you can dogfood the new guidelines. Jon was mentioning this could use another pair of eyes for validation last week.

PR #1178, opened 2024-04-30
Add step to developer-guide to use ppe in msft-learn-mock

ocallesp:


jonsequitur:
What is the use case for this?

ocallesp:
> What is the use case for this?

eliminates the need to manually copy and paste URLs from PPE to the mock server. It is primarily intended for internal developers.

ocallesp:
Given that access to PPE requires explicit authorization, it might not be practical to implement this for all users.

Issue #1192, opened 2024-05-15
exception output is not displayed


jonsequitur:
![image](https://github.com/dotnet/try/assets/547415/88b13460-fa2a-4400-bbd0-fcb9ef1527e2)

AbhitejJohn:
@ocalles : Can you take a look at this with priority please?

ocallesp:
Taking look at it today

PR #1196, opened 2024-05-17
Print exceptions as errors

ocallesp:
image

### DIFF
--- a/src/Microsoft.TryDotNet.WasmRunner/CodeRunner.cs
+++ b/src/Microsoft.TryDotNet.WasmRunner/CodeRunner.cs
@@ -77,7 +77,6 @@ public class CodeRunner
             }
             catch (Exception e)
             {
-                onOutput(e.ToString());
                 switch (e.InnerException)
                 {
                     case TypeLoadException:
--- a/src/Microsoft.TryDotNet.WasmRunner/CodeRunner.cs
+++ b/src/Microsoft.TryDotNet.WasmRunner/CodeRunner.cs
@@ -77,6 +77,7 @@ public class CodeRunner
             }
             catch (Exception e)
             {
+                onOutput(e.ToString());
                 switch (e.InnerException)
                 {
                     case TypeLoadException:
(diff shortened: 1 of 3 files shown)


src/Microsoft.TryDotNet.WasmRunner/CodeRunner.cs
```diff
@@ -77,6 +77,8 @@ private static  Assembly LoadFromBase64EncodedString(string base64EncodedAssembl
             }
             catch (Exception e)
             {
+                onError(e.ToString());
+                onOutput(e.ToString());
```
AbhitejJohn, line 81: I'm not super familiar with this code - it is weird though that we have to call both onError and onOutput for an exception?

AbhitejJohn:
Tagging the issue here: #1192

AbhitejJohn (commented):
Is the error message and details the same as what the current service puts out?

src/Microsoft.TryDotNet.WasmRunner/CodeRunner.cs
colombod: I agree with @AbhitejJohn point. but the exception is wired on the run result, i think we are doing the wrong thing after the runner has produced the reunResult, probably in the integration part

ocallesp: Just to clarify, the changes on this pr is only using onOutput(e.ToString());  
onError() was deleted.

@colombod do you still think the same ?

AbhitejJohn: @ocallesp : This still seem like the wrong place to make the fix since we aren't really passing the exception details through but just printing it to the output pane as Diego mentions. Can you please take a look at the suggestion above and re-work this? At the moment, this seems like a hack.

colombod:
@ocalles  we still need the suggestion from @AbhitejJohn , the `onError` call in the try catch block of the wasm runner is missing.
The `onError` call makes the error from the exception flow as event while the `RunResult` is a synchronous results that contains all outputs and errors.

Do you have a test that shows that your fix flows into the `CSharpProjectWithWasmRunnerIKernel` and then captures in the trudtonet.js so that customer integration can display in the output pane.

colombod (approved):

PR #1208, opened 2024-05-30
css

colombod:
css: Update body dimensions in index.css

The commit updates the width and height properties of the body element in index.css to "auto" instead of fixed values. This change allows for a more responsive layout.

### DIFF
--- a/src/microsoft-trydotnet-editor/src/index.css
+++ b/src/microsoft-trydotnet-editor/src/index.css
@@ -1,5 +1,10 @@
-body {
+.body {
 	width: auto;
 	height: auto;
 	border: 1px solid #ccc;
+	margin: 0;
+}
+
+.monaco-editor {
+	padding: 0.5rem;
 }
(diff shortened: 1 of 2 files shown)


ocallesp (approved):

src/microsoft-trydotnet-editor/src/index.css
```diff
@@ -1,5 +1,5 @@
 body {
-	width: 800px;
-	height: 600px;
+	width: auto;
+	height: auto;
 	border: 1px solid #ccc;
```
jwmunn, line 4: Suggestion:
If you update this to add `margin: 0;` and then add padding to `.monaco-editor` then it will get rid of that white border in dark mode.

```
.body {
	width: auto;
	height: auto;
	border: 1px solid #ccc;
    margin: 0;
 }
 
.monaco-editor {
   padding: 0.5 rem; // 8px
}
```

**Before**
image


**After**
image

colombod: thank you!!

Issue #1198, opened 2024-05-23
Sizing issues of the IFrame containing the interactive editor


ocallesp:
![image](https://github.com/dotnet/try/assets/510598/9e166aa1-8d55-49ef-aa1d-0960fed4d52d)

This is needed in order to improve the UI.

In production it looks seamless
![image](https://github.com/dotnet/try/assets/510598/9b771c03-1531-4921-9657-e8b1ec875f34)

however with the new version we can see some unnecessary overflow
![image](https://github.com/dotnet/try/assets/510598/8108c469-25b9-4eaa-b924-f00d1c5b1a3c)

AbhitejJohn:
Additional context: this is specific to the marketing page.

ocallesp:
https://dotnetwebsite.azurewebsites.net/en-us/  

or for pr changes 

https://dotnetwebsite-dev-pr-6496.azurewebsites.net/

jonsequitur:
The editor can be found here: https://dotnet.microsoft.com/en-us/languages

ocallesp:
The fix should go in the host

jonsequitur:
The element that needs to be sized is here: 
![image](https://github.com/dotnet/try/assets/547415/a003ae15-4602-413a-a926-b32abe70c652)

It should be sized using the Monaco API, which is exposed by trydotnet.js.

jonsequitur:
https://microsoft.github.io/monaco-editor/typedoc/interfaces/editor.IEditor.html#layout.layout-1

danzhu54:
When does the editor set the size by default?
It seems to be called twice: once when it initializes and then again when I set the size with `setSize`
![image](https://github.com/dotnet/try/assets/33338258/37d8912f-8758-444f-a93c-4c0d59380113)

Is it possible for me to pass in my size preferences to the editor via `createSessionWithProjectAndOpenDocument`?

ocallesp:
@danzhu54 you cannot pass the size of the editor via `createSessionWithProjectAndOpenDocument`

PR #1210, opened 2024-06-03
Hide scroll bar in iframe

ocallesp:
https://github.com/dotnet/try/issues/1198

hideScrollBar

Thanks @danzhu54  for the fix suggestion

### DIFF
--- a/src/microsoft-trydotnet-editor/src/index.css
+++ b/src/microsoft-trydotnet-editor/src/index.css
@@ -2,6 +2,7 @@ body {
 	width: auto;
 	height: auto;
 	margin: 0;
+	overflow: hidden;
 }
 
 .monaco-editor {


src/microsoft-trydotnet-editor/src/index.css
```diff
@@ -2,6 +2,7 @@ body {
 	width: auto;
 	height: auto;
 	margin: 0;
+	overflow: hidden;
```
jonsequitur, line 5: If the content is sized correctly, I would think we shouldn't need to do this.

ocallesp: It seems that the issue is not in the content.
Here is the animated gif showing the behavior WITHOUT 'overflow: hidden'
ScrollBar

ocallesp:
@jwmunn does this change break something on Learning ?

src/microsoft-trydotnet-editor/src/index.css
jonsequitur: Let's find some time to discuss.

jonsequitur:
We should close this PR and take a different approach. The editor's API can be called directly by the host page and that's going to be more flexible than one-size-fits-all attempts to remedy this with CSS.

danzhu54:
@jonsequitur will you expose a setSize function on your API instead? I believe this change will resolve all editor overflow issues. Learn is facing the same issue:
image

colombod:
Who is the owner of the scroll bars? is it the iframe or the div inside?

colombod:
This is what I see:
The content of the iframe is bigger than the iframe:
image
the element with class monaco-editor is se to this the size 1245px X 839px in the style sheet (side effect of calling the layout code)

this is the iframe size : 
image

and this is the actual size of the div 
image

jonsequitur:
Let's move this discussion to the issue.

https://github.com/dotnet/try/issues/1198#issuecomment-2148221113

PR #1225, opened 2024-07-11
Add a method to set editor size from container

ocallesp:
https://github.com/dotnet/try/issues/1198


src/microsoft-trydotnet/src/internals/monacoTextEditor.ts
```diff
@@ -113,6 +113,12 @@ export class MonacoTextEditor implements ITrydotnetMonacoTextEditor {
         this.editorApimessageBus.post(request);
     }
 
+    public fitToContainer(container: HTMLElement): void {
+        const offsetWidth = 2;
+        const offsetHeight = 2;
+        this.setSize({ width: container.clientWidth - offsetWidth, height: container.clientHeight - offsetHeight})
```
ocallesp, line 119: @colombod it looks like this method belongs to the project microsoft-trydotnet.

PR #1238, opened 2024-10-16
Create devcontainer.json

mustafacco7:


jonsequitur:
There's already a devcontainer in the repo: https://github.com/dotnet/interactive/blob/main/.devcontainer/devcontainer.json

Please open an issue if there's a specific feature you think is missing.

PR #1248, opened 2025-01-23
[DO NOT MERGE] Install sdk 9 on agent

ocallesp:
Validating a fix

### DIFF
--- a/azure-pipelines-CI.yml
+++ b/azure-pipelines-CI.yml
@@ -132,7 +132,10 @@ extends:
                 skipTests: $(SkipTests)
                 buildArgs: $(_BuildArgs)
                 additionalArgs: $(additionalWindowsArgs)
-
+            - task: UseDotNet@2
+              inputs:
+                packageType: sdk
+                useGlobalJson: true
       - template: /eng/common/templates-official/jobs/jobs.yml@self
         parameters:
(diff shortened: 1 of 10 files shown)


azure-pipelines-CI.yml
```diff
@@ -126,13 +126,17 @@ extends:
               - name: _BuildArgs
                 value: /p:SignType=$(_SignType)
             steps:
+            - task: UseDotNet@2
```
AbhitejJohn, line 129: We might want to put this in the linux leg below as well. Would be easier if it was part of a common task list we can stick in all the jobs for all the other OSs.

azure-pipelines.yml
```diff
@@ -77,13 +77,17 @@ stages:
           - name: _BuildArgs
             value: /p:SignType=$(_SignType)
         steps:
+        - task: UseDotNet@2
+          displayName: Install SDK 9.0.102
+          inputs:
+            packageType: sdk
+            version: 9.0.102
```
AbhitejJohn, line 84: We might want to put this in the linux leg below as well. Would be easier if it was part of a common task list we can stick in all the jobs for all the other OSs.

AbhitejJohn (approved):

PR #1260, opened 2025-02-20
fix: install npm separately

dalehhirt:
NodeJS package removed npm sometime in the past month.


AbhitejJohn (approved):

AbhitejJohn:
Merging without waiting on checks here since I don't think any of these checks will even exercise this code change.

